PR #11, opened 2015-09-08
Django 1.8

kostko:


requirements-readthedocs.txt
```diff
@@ -6,21 +6,20 @@ Sphinx==1.2.3
...
-git+https://github.com/wlanslovenija/django-overextends.git@d70cdf44bff3fd130a0a64f4f6836ca3b8ea030f#egg=django_overextends-wlanslovenija
+--allow-all-external git+https://github.com/wlanslovenija/django-datastream.git@fe61aace1b8c8131ae914b067f9f2a200fedbeec#egg=django_datastream-master
+django-guardian==1.3
+git+https://github.com/wlanslovenija/django-missing.git@028bb5a94e82a0f290cbd6fb85edf7163fcdf05c#egg=django_missing-master
+git+https://github.com/wlanslovenija/django-overextends.git@83d29ef10c781bb751e0cf6f55411c9f44a848dd#egg=django_overextends-wlanslovenija
 django-phonenumber-field==0.5
 django_polymorphic==0.7.1
-git+https://github.com/wlanslovenija/django-registration.git@670779dce3b81fabc7cc2394db1ddd45ffcc4dc7#egg=django_registration-wlanslovenija
-django-sekizai==0.8.1
+git+https://github.com/ubernostrum/django-registration.git@6a953b0044a7e82fca0f9ce3b7146a0dffec3a51#egg=django_registration-master
```
mitar, line 21: Ohhh. I think I have changed `django-registration` quite a bit from the original package. Not sure if just switching a package can really work. Please verify this. I assume we will have to update the whole registration code anyway. To custom Django user model, registration, login and logout notifications based on official signals, and even OAuth support (both as a consumer and service).

kostko, line 21: Yes I know, but I had to update it as the used version was so old it didn't work anymore on Django 1.8.

kostko, line 21: I've ported the things over to the new version, which uses generic views instead of the previously used "backend" class. There were some other fixes required due to changes in the authentication models and forms in Django 1.8.

PR #12, opened 2015-11-22
add TP-Link WDR3600

pberndro:


nodewatcher/modules/devices/tplink/wdr3600.py
```diff
@@ -0,0 +1,106 @@
...
+            cpu_tagged=True,
+            vlans=16,
+        )
+    ]
+    ports = [
+        cgm_devices.SwitchedEthernetPort(
+            'wan0', "Wan0",
+            switch='sw0',
+            vlan=2,
+            ports=[0, 1],
```
kostko, line 55: Did you check this switch configuration?

nodewatcher/modules/devices/tplink/wdr3600.py
```diff
@@ -0,0 +1,106 @@
...
+    profiles = {
+        'openwrt': {
+            'name': 'TLWDR3600',
+            'files': [
+                'openwrt-ar71xx-generic-tl-wdr3600-v1-squashfs-factory.bin'
+            ]
+        }
+    }
+
+# Register the TP-Link WDR3600 device
```
kostko: Please add a dot at the end of any new comments (code style).

kostko:
Also, you need to modify `tplink/__init__.py` so that the new device descriptor module is properly imported.

nodewatcher/modules/devices/tplink/wdr3600.py
pberndro, line 55: yes, the switch config is the same like tp-link wdr4300.
https://wiki.openwrt.org/toh/tp-link/tl-wdr3600#switch_ports_for_vlans

nodewatcher/modules/devices/tplink/wdr3600.py
pberndro: i send a new commit with the dot at end of line, also in the template for tp-link wdr4300 the dot is missing at the same line. i copy the file from wdr4300.

kostko: Yes this is the new code style, so old comments are still not like that and will be gradually fixed.

PR #13, opened 2015-11-22
add TP-Link WDR3600

pberndro:


kostko:
This is not ok, make a single pull request with all the changes required for this device.

mitar:
You can just add to the branch and pull request will be automatically updated.

kostko:
The problem is that this PR was made through the GitHub website, so this can't be done AFAIK. He has opened #15 with a proper PR.

mitar:
It can. He just clones repository locally and goes to `patch-1` branch and continues working there.

mitar:
Or, he goes to the branch in his fork on GitHub, and continues editing there.

The only downside of using GitHub is that for every change to one file there will be one commit. You cannot change multiple files at the same time. In this case he is changing only one file anyway, so it should work.

kostko:
No, he had to edit other files as well as to my comments in #12. Oh, so he can also access that web-created branch via git, ok then it could be done.

pberndro:
thank you guys for support, i was confused by the web ui.
at the cli it was done in 1 minute.

PR #15, opened 2015-11-22
add tp-link wdr3600v1

pberndro:


mitar:
Ah, what I mess you made here. ;-) But OK, you are learning.

One more thing. It is best that you create locally a branch just for your pull request. Changing things directly in `development` branch of your fork can bring you issues when your branch will diverge from ours. For example, if we decide to merge your pull request differently with a different commit. Then your `development` branch will be forever different and every new future pull request will contain old  commits which will be preventing them from being merged.

kostko:
Merged with some updates (the profile needed to be changed) in 1fd42ca7641e7036866db35702eabbd319b15170.

PR #17, opened 2016-03-12
API v2

kostko:
See [#1268](https://dev.wlan-si.net/ticket/1268)


nodewatcher/core/registry/fields.py
```diff
@@ -616,8 +616,6 @@ def get_queryset(self):
                     db = self._db or django_db.router.db_for_read(self.model, instance=self.instance)
                     empty_strings_as_null = django_db.connections[db].features.interprets_empty_strings_as_nulls
                     qs = super(RelatedManager, self).get_queryset()
-                    # Disable polymorphic queries as they can be very expensive in this case.
-                    qs = qs.non_polymorphic()
```
kostko: This needs to be removed, otherwise you will only get top-level items. I am not sure whether we can make it perform any better as the previous comment indicated that this is a major performance hit.

nodewatcher/core/registry/frontend.py
```diff
@@ -0,0 +1,37 @@
...
+        '%sRegistryRootViewSet' % (model.__name__),
+        (views.RegistryRootViewSetMixin, drf_viewsets.ReadOnlyModelViewSet),
+        {
+            'queryset': model.objects.all(),
+            'serializer_class': serializer,
+        }
+    )
+
+    # Register the viewset with the API router.
+    api.router.register(r'registry/%s' % model._meta.object_name.lower(), viewset)
```
kostko: Should we have all top-level registry endpoints under `registry/` or should they be at root directly?

kostko:
Currently, the API does not yet support filtering. For an example output, see [this query](http://127.0.0.1:8000/api/v2/registry/node/?fields=config%7Ccore.location&fields=config%7Ccore.general&fields=config%7Ccore.routerid&fields=config%7Ccore.interfaces&limit=1).

nodewatcher/core/frontend/urls.py
```diff
@@ -12,4 +14,6 @@
     # Frontend components auto-discovery
     components.pool.discover_components()
 
-    urlpatterns = components.pool.get_urls()
+    urlpatterns = components.pool.get_urls() + [
+        urls.url(r'^api/v2/', urls.include(api.router.urls)),
```
mitar: I would move this to `nodewatcher/urls.py`, so that it is not part of frontend.

nodewatcher/urls.py
```diff
@@ -11,14 +11,14 @@
...
-    urls.url(r'^api/', urls.include(api.v1_api.urls, namespace='api', app_name='api')),
-
     # Django admin interface.
     urls.url(r'^admin/', urls.include(admin.site.urls)),
 
     # Frontend.
     urls.url(r'^', urls.include(frontend_urls)),
+
+    # Legacy API.
+    urls.url(r'^api/', urls.include(api.v1_api.urls, namespace='api', app_name='api')),
```
mitar: And leave API above the admin.

nodewatcher/core/registry/frontend.py
mitar: I would keep it at top level.

nodewatcher/core/frontend/urls.py
kostko: Hm, but the whole `api` is part of the frontend? Why should it be separate?

mitar: Because to me it is not part of frontend. :-) API can be used by mobile apps, for example. While our `frontend` app is our "web frontend" in fact.

mitar: So to me frontend is this particular Django component system we have and display stuff.

kostko: But why is then `api` under `nodewatcher.core.frontend.api` and not `nodewatcher.core.api`?

mitar: True. Move it. :-)

mitar:
So some features we had in Tastypie and I think we should keep them, so we should check:
- [x] previous URL if limit=10, offset=5 should exist
- [x] adding cors headers:
  - `response['Access-Control-Allow-Origin'] = '*'`
  - `response['Access-Control-Allow-Headers'] = 'Content-Type'`
  - `response['Access-Control-Allow-Methods'] = 'GET, HEAD, OPTIONS'`
  - `response['Access-Control-Max-Age'] = 60 * 60 # seconds, 1 hour`
- [ ] ~~use ujson, and optimization for gis data which provides `__json__` to directly provide how it should be serialized~~
- [x] `limit=0` should give all results (which in fact mean, should limit results to some internal maximum per page (5000?), while default page limit should be something smaller (100?))
- [ ] ~~display correct schema view for (so human/computer readable)~~
  - [ ] ~~click on "options" should have some meaningful description~~
  - [x] click on get JSON should work (currently you get `Registration point 'node.format' does not exist!`)
  - [x] click on get API should work (currently you get `Registration point 'node.format' does not exist!`)
- [x] prevent access to sensitive data, both for display and filtering
- [ ] ~~allow access to sensitive data based on guardian permissions~~
- [ ] ~~support `global_filter` concept (so that you can filter over some concept of all fields at the same time, so I would make it so that based on which fields you are selecting, you can filter over their values, conceptually, whatever is the JSON output, you can filter over that string of the output)~~
- [ ] ~~have both total count and non-filtered count be in the output (at least when using global filter, but in theory we could use datatables to filter per column as well)~~
- [ ] ~~I see that in listing view there is no URI to the detail view for each result, Tasypie had that, and we are using it in datastream (which is not part of this), but we might want to keep it? on the other hand, this would be extra data which is often not needed~~
- [x] `fields=config__core.project` should return a primary key for project
- [x] `fields=config__core.project__project` should return a whole project subdocument
- [x] `fields=config__core.project__project.name` should return a subdocument with only name
- [x] better names and not `Node Registry Root List` and `Node Registry Root Instance`
- [ ] ~~better breadcrumb, I would suggest that it is `Root / Node List / <to string of node>`~~
- [x] integrate the API explorer with the nodewatcher
- [x] `_metadata` and `_type` probably should not be included by default either, no?
- [ ] ~~an extra format which outputs the output (but input stays the same) into NetJSON format~~
- [x] support for filtering
- [x] support for sorting
- [x] show id fields as `@id` in documents
- [x] use `__` instead of `|`
- [ ] ~~optimize speed of conversion between data and JSON~~
- [x] create tests against API, especially for all more complicated query types, so that we can detect issues when we will later on try to optimize things
- [x] add all JSON-LD meta fields, but not necessary all referenced secondary endpoints (vocabulary)
- [x] filter of all nodes on rectangle gis area (will be used for maps)
- [x] support complex filters (with _and_, _or_, and _not_)

mitar:
But it looks awesome!

mitar:
Maybe we can for now just prevent access to sensitive data and later on open it. Because we do not provide OAuth access to API anyway. Or do we?

mitar:
Benchmarking serialization of a GIS field (`core.location` contains one geolocation, a point) using `ab` tool (`-n 10`). API response contains 1074 results, each containing one geolocation. All queries were the same, including all having extra GeoJSON field.

Old Tastypie v1 API with UJSON and our optimization:

```
Document Path:          /api/v1/node/?format=json&limit=0&fields=location__
Document Length:        276817 bytes

Concurrency Level:      1
Time taken for tests:   17.086 seconds
Complete requests:      10
Failed requests:        0
Total transferred:      2771370 bytes
HTML transferred:       2768170 bytes
Requests per second:    0.59 [#/sec] (mean)
Time per request:       1708.604 [ms] (mean)
Time per request:       1708.604 [ms] (mean, across all concurrent requests)
Transfer rate:          158.40 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:  1443 1708 225.9   1682    2260
Waiting:     1416 1692 229.6   1652    2251
Total:       1443 1709 225.9   1682    2260

Percentage of the requests served within a certain time (ms)
  50%   1682
  66%   1695
  75%   1738
  80%   1831
  90%   2260
  95%   2260
  98%   2260
  99%   2260
 100%   2260 (longest request)
```

v2 API default JSON:

```
Document Path:          /api/v2/node/?format=json&limit=0&fields=config|core.location
Document Length:        293849 bytes

Concurrency Level:      1
Time taken for tests:   84.605 seconds
Complete requests:      10
Failed requests:        0
Total transferred:      2940230 bytes
HTML transferred:       2938490 bytes
Requests per second:    0.12 [#/sec] (mean)
Time per request:       8460.512 [ms] (mean)
Time per request:       8460.512 [ms] (mean, across all concurrent requests)
Transfer rate:          33.94 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:  7593 8460 817.1   8352    9959
Waiting:     7581 8442 819.8   8332    9950
Total:       7593 8460 817.1   8352    9959

Percentage of the requests served within a certain time (ms)
  50%   8352
  66%   8419
  75%   8421
  80%   9830
  90%   9959
  95%   9959
  98%   9959
  99%   9959
 100%   9959 (longest request)
```

v2 API [UJSON](https://github.com/wlanslovenija/drf-ujson-renderer) 1.35:

```
Document Path:          /api/v2/node/?format=json&limit=0&fields=config|core.location
Document Length:        293849 bytes

Concurrency Level:      1
Time taken for tests:   86.984 seconds
Complete requests:      10
Failed requests:        0
Total transferred:      2940230 bytes
HTML transferred:       2938490 bytes
Requests per second:    0.11 [#/sec] (mean)
Time per request:       8698.397 [ms] (mean)
Time per request:       8698.397 [ms] (mean, across all concurrent requests)
Transfer rate:          33.01 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       1
Processing:  8075 8698 894.2   8389   10975
Waiting:     8045 8681 895.7   8377   10964
Total:       8075 8698 894.2   8390   10975

Percentage of the requests served within a certain time (ms)
  50%   8390
  66%   8747
  75%   8878
  80%   9272
  90%  10975
  95%  10975
  98%  10975
  99%  10975
 100%  10975 (longest request)
```

With our optimization of using `__json__`:

```
Document Path:          /api/v2/node/?format=json&limit=0&fields=config|core.location
Document Length:        293845 bytes

Concurrency Level:      1
Time taken for tests:   78.629 seconds
Complete requests:      10
Failed requests:        0
Total transferred:      2940190 bytes
HTML transferred:       2938450 bytes
Requests per second:    0.13 [#/sec] (mean)
Time per request:       7862.865 [ms] (mean)
Time per request:       7862.865 [ms] (mean, across all concurrent requests)
Transfer rate:          36.52 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:  7241 7863 536.0   7903    8812
Waiting:     7225 7845 538.8   7870    8797
Total:       7241 7863 536.0   7903    8812

Percentage of the requests served within a certain time (ms)
  50%   7903
  66%   8021
  75%   8295
  80%   8444
  90%   8812
  95%   8812
  98%   8812
  99%   8812
 100%   8812 (longest request)
```

kostko:
> id fields probably do not have to be shown in registry objects?

Yes they need to be, because there exist relations between some objects. For example the `device` of an interface references the id of another interface. Perhaps we could convert these to JSON references or something?

> support for filtering

So filtering is currently not shown in the DRF API browser, but it is implemented by using conditions as query arguments. For example: `config|core.general|name=home`

mitar:
OK, I made benchmarks and there are terrible. So, v1 is much much more faster. For some reason also the measurements fluctuate a lot. The relative difference for v2 API calls were all up and down between each other between multiple calls so the differences above are not really meaningful. Sometimes one is faster, then another.

The conclusions are:
- our `__json__` optimization did not really improve much in comparison with a similar optimization we did in Tastypie, maybe because something else is slowing down things
- we should see what is a difference between queries done for v2 and v1, number of them, and type of them

mitar:
[Limiting to project name](http://127.0.0.1:8000/api/v2/node/?fields=config|core.location&fields=config|core.project|project.name&config|core.project|project.name=Test) does not seem to work.

mitar:
> Yes they need to be, because there exist relations between some objects. For example the device of an interface references the id of another interface. Perhaps we could convert these to JSON references or something?

I think we should never display and `id` because this is an internal thing of how we have data represented in the database. At least not directly as it is. I think we have [two options](https://www.mnot.net/blog/2011/11/25/linking_in_json):
- going full JSON-lD and start using also `@id` for our ID field, which probably has to be an URL
- start using `$ref` to reference other subdocuments

I would in fact propose that we do both. It seems Django Rest Framework [already supports a form of this](http://www.django-rest-framework.org/tutorial/5-relationships-and-hyperlinked-apis/#hyperlinking-our-api) by using URLs instead of ids. I would propose that instead of using `url` field, we start using `@id` field to reference the document itself. But the content should be the URL of the document.

And for references to not yet expanded subdocument I would use `{$ref: <value of the @id field>}`.

There are another two cases:
- documents which have their own high-level end-point (currently only `node`)
- documents which are embedded somewhere deep in (devices)

For the first I would just use its absolute URL, for the second I would use simply what [JSON reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-00) is proposing, a JSON pointer as an anchor to internals of that document which does have an absolute URL.

Concrete examples, currently we [return this](http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/?fields=config|core.location&fields=config|core.project):

``` json
{
    "uuid": "000c3d9d-5d7c-492a-8d3b-bb8b922ae35f",
    "config": {
        "core.location": {
            "id": 644,
            "timezone": "Europe/Ljubljana",
            "address": "San Juan, Puerto Rico",
            "city": "Ljubljana",
            "country": "SI",
            "geolocation": {
                "type": "Point",
                "coordinates": [
                    -66.073756,
                    18.378287
                ]
            },
            "altitude": 0.0,
            "@type": "Location"
        },
        "core.project": {
            "id": 644,
            "project": 2,
            "@type": "Project"
        }
    }
}
```

I propose:

``` json
{
    "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/",
    "@type": "Node",
    "uuid": "000c3d9d-5d7c-492a-8d3b-bb8b922ae35f",
    "config": {
        "core.location": {
            "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/#/config/core.location",
            "@type": "Location",
            "timezone": "Europe/Ljubljana",
            "address": "San Juan, Puerto Rico",
            "city": "Ljubljana",
            "country": "SI",
            "geolocation": {
                "type": "Point",
                "coordinates": [
                    -66.073756,
                    18.378287
                ]
            },
            "altitude": 0.0
        },
        "core.project": {
            "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/#/config/core.project",
            "@type": "Project",
            "project": {
              "$ref": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/#/config/core.project/project"
            }
        }
    }
}
```

Another [example](http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/?fields=config|core.interfaces):

``` json
{
    "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/",
    "@type": "Node",
    "uuid": "000c3d9d-5d7c-492a-8d3b-bb8b922ae35f",
    "config": {
        "core.interfaces": [
            {
                "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/#/config/core.interfaces/0",
                "@type": "WifiRadioDevice",
                "enabled": true,
                "wifi_radio": "wifi0",
                "protocol": "ieee-80211bg",
                "channel_width": "ht20",
                "channel": "ch11",
                "antenna_connector": null,
                "tx_power": null,
                "ack_distance": null
            },
            {
                "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/#/config/core.interfaces/1",
                "@type": "WifiInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mode": "mesh",
                "essid": "mesh.wlan-si.net",
                "bssid": "02:CA:FF:EE:BA:BE",
                "bitrates_preset": null,
                "bitrates": [],
                "isolate_clients": true,
                "uplink": false,
                "device": {
                  "$ref": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/#/config/core.interfaces/0"
                },
                "connect_to": null
            },
            {
                "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/#/config/core.interfaces/2",
                "@type": "EthernetInterface",
                "enabled": true,
                "routing_protocols": [],
                "eth_port": "wan0",
                "uplink": true,
                "mac_address": null
            },
            {
                "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/#/config/core.interfaces/3",
                "@type": "EthernetInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "eth_port": "lan0",
                "uplink": false,
                "mac_address": null
            },
            {
                "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/#/config/core.interfaces/4",
                "@type": "TunneldiggerInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mac": "00:ff:58:d3:aa:91",
                "server": {
                  "$ref": "http://127.0.0.1:8000/api/v2/server/1/"
                },
                "uplink_interface": null
            },
            {
                "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/#/config/core.interfaces/5",
                "@type": "TunneldiggerInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mac": "00:ff:27:30:8d:c9",
                "server": {
                  "$ref": "http://127.0.0.1:8000/api/v2/server/2/"
                },
                "uplink_interface": null
            }
        ]
    }
}
```

What do you think? My only issue is that those internal ids for interfaces might be changing as elements are moving around in the array, despite being in fact the same element as before, just on a new position. Those internal ids can be seen as valid at a given moment anyway (because they do depend on content anyway) and that if you want permalinks you should move a document to its own top-level API entry.

mitar:
I used [Query Inspector](https://github.com/dobarkod/django-queryinspect) and those are results:

```
web_1           | [INFO/qinspect.middleware] [SQL] 24 queries (3 duplicates), 77 ms SQL time, 1610 ms total request time
web_1           | [14/Mar/2016 01:40:12] "GET /api/v1/node/?format=json&limit=0&fields=location__ HTTP/1.1" 200 276817
web_1           | [INFO/qinspect.middleware] [SQL] 3 queries (0 duplicates), 25 ms SQL time, 7142 ms total request time
web_1           | [14/Mar/2016 01:40:30] "GET /api/v2/node/?format=json&limit=0&fields=config%7Ccore.location HTTP/1.1" 200 293849
```

I cannot believe. With v2 we are in fact down to only 3 queries and only 25 ms SQL time (from 77 ms before). This seems great.

But to spend 7 seconds then to convert to JSON? This is crazy! We should just expose PostgreSQL port to JavaScript or something.

kostko:
> Limiting to project name does not seem to work.

Should be fixed now.

kostko:
> for the second I would use simply what JSON reference is proposing, a JSON pointer as an anchor to internals of that document which does have an absolute URL.

Could we just use the reference without prepending the full absolute URL (so something like `#/config/core.interfaces/5`)? As otherwise, the URL would actually be incorrect (you need to specify `fields` to get the proper fields) and it adds extra space to all references.

mitar:
> Could we just use the reference without prepending the full absolute URL

Where? In `$ref` we can, if they are pointing to current document. In `@id` probably not, because ids should probably be globally unique, no?

> As otherwise, the URL would actually be incorrect

I do not think this is a problem. It is not really URL, it is an identifier, a relative one based on an absolute one. So if the main document's ID is `http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/`, then the sub-document's ID is correct. Now, the fact that the ID of the main document is not a complete URL to fetch the data, but it is more an URI. :-) The fact that you have to add extra query parameters is just implementation detail. So yes, we have URIs, not URLs. ;-)

In fact, they are [IRIs](https://www.w3.org/TR/json-ld/#iris).

> it adds extra space to all references.

It seems IRIs (`@id`s) can be relative. Then let's do:

``` json
{
    "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/",
    "@type": "Node",
    "uuid": "000c3d9d-5d7c-492a-8d3b-bb8b922ae35f",
    "config": {
        "core.location": {
            "@id": "#/config/core.location",
            "@type": "Location",
            "timezone": "Europe/Ljubljana",
            "address": "San Juan, Puerto Rico",
            "city": "Ljubljana",
            "country": "SI",
            "geolocation": {
                "type": "Point",
                "coordinates": [
                    -66.073756,
                    18.378287
                ]
            },
            "altitude": 0.0
        },
        "core.project": {
            "@id": "#/config/core.project",
            "@type": "Project",
            "project": {
              "$ref": "#/config/core.project/project"
            }
        }
    }
}
```

``` json
{
    "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/",
    "@type": "Node",
    "uuid": "000c3d9d-5d7c-492a-8d3b-bb8b922ae35f",
    "config": {
        "core.interfaces": [
            {
                "@id": "#/config/core.interfaces/0",
                "@type": "WifiRadioDevice",
                "enabled": true,
                "wifi_radio": "wifi0",
                "protocol": "ieee-80211bg",
                "channel_width": "ht20",
                "channel": "ch11",
                "antenna_connector": null,
                "tx_power": null,
                "ack_distance": null
            },
            {
                "@id": "#/config/core.interfaces/1",
                "@type": "WifiInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mode": "mesh",
                "essid": "mesh.wlan-si.net",
                "bssid": "02:CA:FF:EE:BA:BE",
                "bitrates_preset": null,
                "bitrates": [],
                "isolate_clients": true,
                "uplink": false,
                "device": {
                  "$ref": "#/config/core.interfaces/0"
                },
                "connect_to": null
            },
            {
                "@id": "#/config/core.interfaces/2",
                "@type": "EthernetInterface",
                "enabled": true,
                "routing_protocols": [],
                "eth_port": "wan0",
                "uplink": true,
                "mac_address": null
            },
            {
                "@id": "#/config/core.interfaces/3",
                "@type": "EthernetInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "eth_port": "lan0",
                "uplink": false,
                "mac_address": null
            },
            {
                "@id": "#/config/core.interfaces/4",
                "@type": "TunneldiggerInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mac": "00:ff:58:d3:aa:91",
                "server": {
                  "$ref": "http://127.0.0.1:8000/api/v2/server/1/"
                },
                "uplink_interface": null
            },
            {
                "@id": "#/config/core.interfaces/5",
                "@type": "TunneldiggerInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mac": "00:ff:27:30:8d:c9",
                "server": {
                  "$ref": "http://127.0.0.1:8000/api/v2/server/2/"
                },
                "uplink_interface": null
            }
        ]
    }
}
```

mitar:
(I think that we should also have `@base` next to top-level `@id`, to be completely by the standard, but I think it is good enough.)

mitar:
BTW, I would create another top-level element for servers, no? Those are separate models?

kostko:
Of course. There are quite some non-registry top-level endpoints (servers, allocation pools, projects, etc.).

mitar:
OK. Then I decided well in the example. :-)

Hm, projects are separate? Then we should change this as well?

``` json
        "core.project": {
            "@id": "#/config/core.project",
            "@type": "Project",
            "project": {
              "$ref": "#/config/core.project/project"
            }
        }
```

So is the fact the we have to explicitly resolve this foreign key only because it is a separate model? Or would you require resolving foreign keys between registry items explicitly as well?

mitar:
BTW, why are we using `|` and not simply `__` if `|` more or less works in the same way?

mitar:
OK, doing indexes are hard. We do not have that information in the database. So I propose that we change anchors a bit and use a slightly modified JSON pointer.

``` json
{
    "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/",
    "@type": "Node",
    "uuid": "000c3d9d-5d7c-492a-8d3b-bb8b922ae35f",
    "config": {
        "core.location": {
            "@id": "#/config/core.location",
            "@type": "Location",
            "timezone": "Europe/Ljubljana",
            "address": "San Juan, Puerto Rico",
            "city": "Ljubljana",
            "country": "SI",
            "geolocation": {
                "type": "Point",
                "coordinates": [
                    -66.073756,
                    18.378287
                ]
            },
            "altitude": 0.0
        },
        "core.project": {
            "@id": "#/config/core.project",
            "@type": "Project",
            "project": {
                "$ref": "http://127.0.0.1:8000/api/v2/project/2/"
            }
        }
    }
}
```

``` json
{
    "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f/",
    "@type": "Node",
    "uuid": "000c3d9d-5d7c-492a-8d3b-bb8b922ae35f",
    "config": {
        "core.interfaces": [
            {
                "@id": "#/config/core.interfaces/@123",
                "@type": "WifiRadioDevice",
                "enabled": true,
                "wifi_radio": "wifi0",
                "protocol": "ieee-80211bg",
                "channel_width": "ht20",
                "channel": "ch11",
                "antenna_connector": null,
                "tx_power": null,
                "ack_distance": null
            },
            {
                "@id": "#/config/core.interfaces/@213",
                "@type": "WifiInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mode": "mesh",
                "essid": "mesh.wlan-si.net",
                "bssid": "02:CA:FF:EE:BA:BE",
                "bitrates_preset": null,
                "bitrates": [],
                "isolate_clients": true,
                "uplink": false,
                "device": {
                    "$ref": "#/config/core.interfaces/@123"
                },
                "connect_to": null
            },
            {
                "@id": "#/config/core.interfaces/@453",
                "@type": "EthernetInterface",
                "enabled": true,
                "routing_protocols": [],
                "eth_port": "wan0",
                "uplink": true,
                "mac_address": null
            },
            {
                "@id": "#/config/core.interfaces/@454",
                "@type": "EthernetInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "eth_port": "lan0",
                "uplink": false,
                "mac_address": null
            },
            {
                "@id": "#/config/core.interfaces/@457",
                "@type": "TunneldiggerInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mac": "00:ff:58:d3:aa:91",
                "server": {
                    "$ref": "http://127.0.0.1:8000/api/v2/server/1/"
                },
                "uplink_interface": null
            },
            {
                "@id": "#/config/core.interfaces/@676",
                "@type": "TunneldiggerInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mac": "00:ff:27:30:8d:c9",
                "server": {
                    "$ref": "http://127.0.0.1:8000/api/v2/server/2/"
                },
                "uplink_interface": null
            }
        ]
    }
}
```

But it is unclear how exactly this JSON pointer extension would be specified. Because it is a bit strange. Because or our `@id`s are not nicely relative anymore (we use `@id` == `123`), and then `@123` means to find an array item with `123` as `@id` at a given position in the document. Or we use a relative `@id` like `#/config/core.interfaces/@123`, but then it is unclear how to resolve `../@123` inside the JSON path.

mitar:
I opened a question on the [JSON-LD mailing list](https://lists.w3.org/Archives/Public/public-linked-json/2016Mar/0013.html).

mitar:
OK, a new take, pure JSON-LD:

``` json
{
    "@context": {
        "@base": "http://127.0.0.1:8000/api/v2/node/",
        "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#",
        "config": {
            "@id": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config",
            "@container": "@index"
        },
        "project": {
            "@id": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.project/project",
            "@type": "@id"
        }
    },
    "@type": "Node",
    "@id": "000c3d9d-5d7c-492a-8d3b-bb8b922ae35f",
    "config": {
        "core.location": {
            "@context": {
                "@base": "_:config/core.location/",
                "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.location/"
            },
            "@id": "123",
            "@type": "Location",
            "timezone": "Europe/Ljubljana",
            "address": "San Juan, Puerto Rico",
            "city": "Ljubljana",
            "country": "SI",
            "geolocation": {
                "@context": "http://geojson.org/contexts/geojson-base.jsonld",
                "type": "Point",
                "coordinates": [
                    -66.073756,
                    18.378287
                ]
            },
            "altitude": 0.0
        },
        "core.project": {
            "@context": {
                "@base": "_:config/core.project/",
                "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.project/"
            },
            "@id": "444",
            "@type": "Project",
            "project": "http://127.0.0.1:8000/api/v2/project/2"
        }
    }
}
```

`config` and `project` should not be defined inline inside `@context` (but inside `http://127.0.0.1:8000/api/v2/node/?format=vocab`), but are in the example to show the special definitions needed in their definitions. Also `project` should not be defined at the top-level at all.

Do note that the URL (and `@id`) for the project document is `http://127.0.0.1:8000/api/v2/project/2` and not `http://127.0.0.1:8000/api/v2/project/2/`. We might re-consider the use of trailing slash. Maybe we could decide that in API, for sets we use trailing slash (`http://127.0.0.1:8000/api/v2/project/`), but for particular item we do not (`http://127.0.0.1:8000/api/v2/project/2`). Like directory (multiple items) vs. item itself.

I also do not like the use of `format=vocab`, it makes things long, but maybe this is the best we can do? Is there a query string to force `OPTIONS` (Tastypie had it)? So I realized that schema is more or less linked to a particular document. Alternatively I was thinking we could have `http://127.0.0.1:8000/api/v2/node.vocab/` (but then it is unclear about trailing slash, should it be there or not), because [even Django REST Framework supports such extensions as an alternative to the `format=` query string](http://www.django-rest-framework.org/api-guide/format-suffixes/).

``` json
{
    "@context": {
        "@base": "http://127.0.0.1:8000/api/v2/node/",
        "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#",
        "config": {
            "@id": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config",
            "@container": "@index"
        },
        "server": {
            "@id": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.interfaces/server",
            "@type": "@id"
        },
        "device": {
            "@id": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.interfaces/device",
            "@type": "@id"
        }
    },
    "@type": "Node",
    "@id": "000c3d9d-5d7c-492a-8d3b-bb8b922ae35f",
    "config": {
        "core.interfaces": [
            {
                "@context": {
                    "@base": "_:config/core.interfaces/",
                    "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.interfaces/"
                },
                "@id": "123",
                "@type": "WifiRadioDevice",
                "enabled": true,
                "wifi_radio": "wifi0",
                "protocol": "ieee-80211bg",
                "channel_width": "ht20",
                "channel": "ch11",
                "antenna_connector": null,
                "tx_power": null,
                "ack_distance": null
            },
            {
                "@context": {
                    "@base": "_:config/core.interfaces/",
                    "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.interfaces/"
                },
                "@id": "213",
                "@type": "WifiInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mode": "mesh",
                "essid": "mesh.wlan-si.net",
                "bssid": "02:CA:FF:EE:BA:BE",
                "bitrates_preset": null,
                "bitrates": [],
                "isolate_clients": true,
                "uplink": false,
                "device": "123",
                "connect_to": null
            },
            {
                "@context": {
                    "@base": "_:config/core.interfaces/",
                    "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.interfaces/"
                },
                "@id": "453",
                "@type": "EthernetInterface",
                "enabled": true,
                "routing_protocols": [],
                "eth_port": "wan0",
                "uplink": true,
                "mac_address": null
            },
            {
                "@context": {
                    "@base": "_:config/core.interfaces/",
                    "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.interfaces/"
                },
                "@id": "454",
                "@type": "EthernetInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "eth_port": "lan0",
                "uplink": false,
                "mac_address": null
            },
            {
                "@context": {
                    "@base": "_:config/core.interfaces/",
                    "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.interfaces/"
                },
                "@id": "457",
                "@type": "TunneldiggerInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mac": "00:ff:58:d3:aa:91",
                "server": "http://127.0.0.1:8000/api/v2/server/1",
                "uplink_interface": null
            },
            {
                "@context": {
                    "@base": "_:config/core.interfaces/",
                    "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.interfaces/"
                },
                "@id": "676",
                "@type": "TunneldiggerInterface",
                "enabled": true,
                "routing_protocols": [
                    "olsr",
                    "babel"
                ],
                "mac": "00:ff:27:30:8d:c9",
                "server": "http://127.0.0.1:8000/api/v2/server/2",
                "uplink_interface": null
            }
        ]
    }
}
```

Similarly, `config`, `server`, and `device` should not be defined inline here.

`device` can be simply `123` because `@base` applies for it as well. If the reference would be to something else then this trick would not work.

Use [JSON-LD Playground](http://json-ld.org/playground/) see how those are parsed.

mitar:
Opened: https://github.com/json-ld/json-ld.org/issues/415

kostko:
I have added tests for everything but JSON-LD. The tests use only registry items defined under `nodewatcher.core` and all other tests (type, project, location, etc.) should live in their respective modules (when they are needed).

I plan on merging this soon to support a list of nodes with some custom fields. The remaining features can be added later.

mitar:
OK, but the features which I didn't move to https://dev.wlan-si.net/ticket/1412 ticket, are those which we need to support current features (like current list of nodes), for example, global search.

We also have to think about how to tell what fields one can search over anyway. So describe the API endpoint.

kostko:
What do you mean, search is already supported in this branch and the data table in this branch already uses it. You can specify an arbitrary field to search by and it does a filter.

The only missing thing in this regard is the "total count" reporting (with and without the applied filter). But since search is just a normal filter, there needs to be a way to specify that some filter(s) are special. I am not sure if this is the best approach.

mitar:
> What do you mean, search is already supported in this branch and the data table in this branch already uses it. You can specify an arbitrary field to search by and it does a filter.

No, the filter in Tastypie implementation searches over multiple fields at the same time, it is a global filter. (So it filters on some fields with **or** between conditions.)

Is this supported in new version?

kostko:
Oh, I see. No, this is not currently supported, but I can easily add it. The question is just what kind of syntax should we use for specifying such queries?

mitar:
BTW. I think we should really do the rest of the list above. If we merge now, I think it will take a long before we do other stuff. And then we will hack for those cases.

> Oh, I see. No, this is not currently supported, but I can easily add it. The question is just what kind of syntax should we use for specifying such queries?

In current API there is "global filter". Maybe each registry item could specify which fields in the item should be used (indexed) for global search (this could be used also for some in the future global search). And then if in the API you selected some fields, and those fields are among global search fields, then when you do "global filter" then you search over them.

API could be maybe something like `__filter__=` or `_filter=`? Or just `global_filter=`? How you currently have for maintainers?

mitar:
If project is embedded, then API should return:

``` json
        "core.project": {
            "@context": {
                "@base": "_:config/core.project/",
                "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.project/"
            },
            "@id": "444",
            "@type": "Project",
            "project": {
              "@context": {
                  "@base": "http://127.0.0.1:8000/api/v2/project/",
                  "@vocab": "http://127.0.0.1:8000/api/v2/project/?format=vocab#"
              },
              "@id": "2",
              "name": "Slovenija"
          }
        }
```

mitar:
So if project is embedded, whole first output is:

``` json
{
    "@context": {
        "@base": "http://127.0.0.1:8000/api/v2/node/",
        "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#",
        "config": {
            "@id": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config",
            "@container": "@index"
        },
        "project": {
            "@id": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.project/project",
            "@type": "@id"
        }
    },
    "@type": "Node",
    "@id": "000c3d9d-5d7c-492a-8d3b-bb8b922ae35f",
    "config": {
        "core.location": {
            "@context": {
                "@base": "_:config/core.location/",
                "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.location/"
            },
            "@id": "123",
            "@type": "Location",
            "timezone": "Europe/Ljubljana",
            "address": "San Juan, Puerto Rico",
            "city": "Ljubljana",
            "country": "SI",
            "geolocation": {
                "@context": "http://geojson.org/contexts/geojson-base.jsonld",
                "type": "Point",
                "coordinates": [
                    -66.073756,
                    18.378287
                ]
            },
            "altitude": 0.0
        },
        "core.project": {
            "@context": {
                "@base": "_:config/core.project/",
                "@vocab": "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.project/"
            },
            "@id": "444",
            "@type": "Project",
            "project": {
              "@context": {
                  "@base": "http://127.0.0.1:8000/api/v2/project/",
                  "@vocab": "http://127.0.0.1:8000/api/v2/project/?format=vocab#"
              },
              "@id": "2",
              "name": "Slovenija"
          }
        }
      }
    }
}
```

It seems that this expansion works well. The first version (only with id) json-ld playground expands into:

``` json
      {
        "@id": "_:config/core.project/444",
        "@type": [
          "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.project/Project"
        ],
        "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.project/project": [
          {
            "@id": "http://127.0.0.1:8000/api/v2/project/2"
          }
        ],
        "@index": "core.project"
      }
```

And if embedded it becomes:

``` json
      {
        "@id": "_:config/core.project/444",
        "@type": [
          "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.project/Project"
        ],
        "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.project/project": [
          {
            "@id": "http://127.0.0.1:8000/api/v2/project/2",
            "http://127.0.0.1:8000/api/v2/project/?format=vocab#name": [
              {
                "@value": "Slovenija"
              }
            ]
          }
        ],
        "@index": "core.project"
      }
```

The whole thing then is:

``` json
[
  {
    "@id": "http://127.0.0.1:8000/api/v2/node/000c3d9d-5d7c-492a-8d3b-bb8b922ae35f",
    "@type": [
      "http://127.0.0.1:8000/api/v2/node/?format=vocab#Node"
    ],
    "http://127.0.0.1:8000/api/v2/node/?format=vocab#config": [
      {
        "@id": "_:config/core.location/123",
        "@type": [
          "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.location/Location"
        ],
        "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.location/address": [
          {
            "@value": "San Juan, Puerto Rico"
          }
        ],
        "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.location/altitude": [
          {
            "@value": 0
          }
        ],
        "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.location/city": [
          {
            "@value": "Ljubljana"
          }
        ],
        "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.location/country": [
          {
            "@value": "SI"
          }
        ],
        "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.location/geolocation": [
          {
            "http://example.com/vocab#coordinates": [
              {
                "@value": -66.073756
              },
              {
                "@value": 18.378287
              }
            ],
            "http://example.com/vocab#type": [
              {
                "@value": "Point"
              }
            ]
          }
        ],
        "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.location/timezone": [
          {
            "@value": "Europe/Ljubljana"
          }
        ],
        "@index": "core.location"
      },
      {
        "@id": "_:config/core.project/444",
        "@type": [
          "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.project/Project"
        ],
        "http://127.0.0.1:8000/api/v2/node/?format=vocab#config/core.project/project": [
          {
            "@id": "http://127.0.0.1:8000/api/v2/project/2",
            "http://127.0.0.1:8000/api/v2/project/?format=vocab#name": [
              {
                "@value": "Slovenija"
              }
            ]
          }
        ],
        "@index": "core.project"
      }
    ]
  }
]
```

See `@index`. In JSON-LD I made it so that instead of us having two-level dict, we have just an array with `@index` which is the config key.

kostko:
> API could be maybe something like `__filter__=` or `_filter=`? Or just `global_filter=`? How you currently have for maintainers?

So we need two parameters, one to specify the fields by which we should filter and the other one to specify the value to filter for.

Maintainers are currently just a `maintainer` parameter. We should probably rename it to something more generic, as in theory, this is not just for nodes but for any kinds of objects. Perhaps it should be `has_permissions` or something.

mitar:
No, the client should not have to configure which field to search over. Or would you like to make this customization?

My idea is that simply with `fields=` you select which fields you want returned. Then, if you do `global_filter=` it searches over those fields you are selected which have some "global filter enabled" flag turned on.

> Perhaps it should be `has_permissions` or something.

Sounds good.

kostko:
> No, the client should not have to configure which field to search over. Or would you like to make this customization?

Yes, the client should configure this, how can the server otherwise know which fields to filter on? I don't think that just using all projected fields would be ok and in some cases it would not even be possible (for example a projection may project an object, which you can't filter on).

mitar:
> Yes, the client should configure this, how can the server otherwise know which fields to filter on?

Because each field or registry item would tell which fields have a reasonable way to search when specified a generic string.

How could client know which of fields have a reasonable text search option?

Hm, maybe because for that report UI generation we will have to know which fields are textual anyway?

mitar:
What I am saying is that "global filter" is to me some extension point different modules can register themselves in. And can say, "if user is interested (fields) into this registry item, this is how you can global search on it."

kostko:
> What I am saying is that "global filter" is to me some extension point different modules can register themselves in. And can say, "if user is interested (fields) into this registry item, this is how you can global search on it."

So you are saying that as soon as a specific registry item is projected, some fields (declared by the registry item developer) would become part of the "global filter"? So for example, projecting anything under `core.project` would automatically add `config:core.project__project__name` to the list of global filters (for example)?

> How could client know which of fields have a reasonable text search option?

Well, the client is already allowed to filter by arbitrary fields, including using text expressions. For example, the client may do `config:core.project__project__name__icontains=foo`.

mitar:
> So for example, projecting anything under core.project would automatically add config:core.project__project__name to the list of global filters (for example)?

Maybe projecting `config:core.project__project__name` would add `config:core.project__project__name` if it is specified as searchable.

mitar:
> Well, the client is already allowed to filter by arbitrary fields, including using text expressions.

Yes, but the question is how it knows which of those queries can it do? I am not talking here about a person, but client program. So user can reason about this. But how can a "global filter" field know which queries to construct in the API request when user (human) types something in?

kostko:
> But how can a "global filter" field know which queries to construct in the API request when user (human) types something in?

The developer of the client-side application will decide which filters to use? The same as it will decide which columns to show or which filters to apply.

mitar:
And what about when user can select which columns to display through UI interface? And datatables display rows. But there is still one field to search over globally.

mitar:
But yes, we can do then something like this:
- you create normal filters
- documents returned from those filters are those which count towards "total count"
- you can additionally add second-layer of filters, which are **or**ed together, those count towards "filtered count"

And it will be developers or UIs responsibility to generate this second-layer of filters somehow.

mitar:
In general, is there a way to create **and** and **or** between filters? Because I was thinking of using that in my UI, but now I do not even believe this is possible through the query language exposed?

kostko:
The registry supports such queries, but they are currently not exposed via the API. If you think of a nice way to pass such query expressions, we can support it.

kostko:
We now have support for complex filter expressions using `filters`.

mitar:
Super! I updated the list above. I think only GIS query is now pending. Everything else has been done!

nodewatcher/core/api/serializers.py
```diff
@@ -18,12 +18,22 @@
...
+    """
+
     def to_representation(self, instance):
         data = super(JSONLDSerializerMixin, self).to_representation(instance)
 
         # Fix primary key field.
+        id_field = None
         if instance._meta.pk.name in data:
-            del data[instance._meta.pk.name]
+            id_field = instance._meta.pk.name
```
mitar, line 32: Hm, if we expose primary key as `@id`, how can you then know over which field you can filter and sort? Because you probably still have to write `uuid` to filter, and not `@id`?

PR #19, opened 2016-05-07
Added TP-Link WR841NDv11 device descriptor.

valentt:


mitar:
Has this been tested? Has this been tested on CC revision we have currently [builders for](https://github.com/wlanslovenija/firmware-core)?

valentt:
Which revision of CC do we use currently? Support for V11 was added in 49228, so if it is not older than 49228 it should work.

mitar:
You can see [here which Docker images we are using](https://github.com/wlanslovenija/wlanslovenija-servers/blob/master/pillars/aggregator/docker/builders.sls).

So for example, `builder-openwrt-vb106cfb-cc-ar71xx` means that this is `b106cfb` revision of [this repository](https://github.com/wlanslovenija/firmware-core/commits/master) which is this change: https://github.com/wlanslovenija/firmware-core/commit/b106cfb0a8f35d1af09a75e02fb245ffef449868

Now you can search repository at that revision and find which CC version we are using: https://github.com/wlanslovenija/firmware-core/blob/b106cfb0a8f35d1af09a75e02fb245ffef449868/openwrt/branches

Which you can see is this one: https://github.com/openwrt/openwrt/commit/88ef6c4258fe7201678c3862d9a4f0da391a5155

From `git-svn-id` it seems this is:

```
git-svn-id: svn://svn.openwrt.org/openwrt/branches/chaos_calmer@49007
```

As 49007 < 49228 we do not have support for the device you are proposing in our firmware.

mitar:
So, you have first to upgrade our firmware, test it out, and then try building image for v11 and test it out.

kostko:
The device descriptor looks good to me, assuming that all other hardware specifics are the same as in v7 and that the profile and image filenames are correct. I think we can go ahead and merge this in, but in order to use it, we need the latest CC builders (I assume these are in wlanslovenija/firmware-core#22). 

Until the new CC revision is in and all the builders have been built, using this device will result in a failed build.

kostko:
Added in 2fb7101e67c457257426fb1e6797fdbf798ec47d.

PR #22, opened 2016-06-03
Survey module

CdavM:
This PR should be used to discuss the implementation of the survey module, which will collect data about node SNR, neighbors and channel allocations.

### DIFF
--- a/nodewatcher/modules/monitor/http/survey/cgm.py
+++ b/nodewatcher/modules/monitor/http/survey/cgm.py
@@ -7,4 +7,4 @@ def clients(node, cfg):
     Register packages needed for reporting the clients.
     """
 
-    cfg.packages.update(['nodewatcher-agent-mod-clients'])
+    cfg.packages.update(['nodewatcher-agent-mod-wireless'])
--- a/nodewatcher/modules/monitor/http/survey/cgm.py
+++ b/nodewatcher/modules/monitor/http/survey/cgm.py
@@ -1,0 +1,10 @@
+from nodewatcher.core.generator.cgm import base as cgm_base
+
+
+@cgm_base.register_platform_module('openwrt', 50)
+def clients(node, cfg):
+    """
+    Register packages needed for reporting the clients.
+    """
+
+    cfg.packages.update(['nodewatcher-agent-mod-clients'])
--- a/nodewatcher/modules/monitor/http/survey/cgm.py
+++ b/nodewatcher/modules/monitor/http/survey/cgm.py
@@ -2,7 +2,7 @@ from nodewatcher.core.generator.cgm import base as cgm_base
 
 
 @cgm_base.register_platform_module('openwrt', 50)
-def clients(node, cfg):
+def wireless(node, cfg):
     """
     Register packages needed for reporting wireless metadata.
     """
--- a/nodewatcher/modules/monitor/http/survey/cgm.py
+++ b/nodewatcher/modules/monitor/http/survey/cgm.py
@@ -4,7 +4,7 @@ from nodewatcher.core.generator.cgm import base as cgm_base
 @cgm_base.register_platform_module('openwrt', 50)
 def clients(node, cfg):
     """
-    Register packages needed for reporting the clients.
+    Register packages needed for reporting wireless metadata.
     """
 
     cfg.packages.update(['nodewatcher-agent-mod-wireless'])
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -21,7 +21,8 @@ class SurveyInfoStreams(ds_models.RegistryRootStreams):
         # The nodewatcher-agent performs a survey once every ~240 monitoring intervals according to
         # https://github.com/wlanslovenija/nodewatcher-agent/blob/master/modules/wireless.c#L362
         # and a monitoring run is performed every 30 seconds according to
-        # https://github.com/wlanslovenija/nodewatcher-agent/blob/73f2b25db2c34ae7d70904bac31379a2243bade0/modules/wireless.c#L437.
+        # https://github.com/wlanslovenija/nodewatcher-agent/blob/73f2b25db2c34ae7d70904bac31379a2243bade0/modules/wireless.c#L437
+        # (master branch at https://github.com/wlanslovenija/nodewatcher-agent/blob/master/modules/wireless.c#L437).
         # So a survey is performed once every two hours, meaning that we use hourly granularity.
         return datastream.Granularity.Hours
 
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -21,11 +21,10 @@ class SurveyInfoStreams(ds_models.RegistryRootStreams):
         return 'monitor.http.survey'
 
     def get_stream_highest_granularity(self):
-        # The nodewatcher-agent performs a survey once every ~240 monitoring intervals according to
-        # https://github.com/wlanslovenija/nodewatcher-agent/blob/master/modules/wireless.c#L362
-        # and a monitoring run is performed every 30 seconds according to
-        # https://github.com/wlanslovenija/nodewatcher-agent/blob/master/modules/wireless.c#L437
-        # So a survey is performed once every two hours, meaning that we use hourly granularity.
+        # The nodewatcher-agent performs a survey once every ~240 monitoring intervals and
+        # a monitoring run is performed every 30 seconds according to the wireless module of
+        #  nodewatcher-agent. So a survey is performed once every two hours, meaning that we
+        #  use hourly granularity.
         return datastream.Granularity.Hours
 
 
@@ -108,9 +107,7 @@ class SurveyInfo(monitor_processors.NodeProcessor):
         if latest_graph != latest_stored_graph:
             # Since a new survey is performed once every two hours, it should be impossible to insert new data more
             # often than once every hour.
-            assert not latest_stored_graph or timezone.now() - streams[0]['latest_datapoint'] >= datetime.timedelta(
-                hours=1,
-            )
+            assert not latest_stored_graph or timezone.now() - streams[0]['latest_datapoint'] >= datetime.timedelta(hours=1)
             # Store the latest graph into datastream.
             context.datastream.survey_topology = SurveyInfoStreamsData(node, latest_graph)
         return context
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -27,12 +27,9 @@ class SurveyInfoStreams(ds_models.RegistryRootStreams):
 
 
 class SurveyInfoStreamsData(object):
-    def __init__(self, node, vertices, edges):
+    def __init__(self, node, neighbor_graph):
         self.node = node
-        self.neighbor_graph = {
-            'v': vertices,
-            'e': edges
-        }
+        self.neighbor_graph = neighbor_graph
 
 
 ds_pool.register(SurveyInfoStreamsData, SurveyInfoStreams)
@@ -84,6 +81,25 @@ class SurveyInfo(monitor_processors.NodeProcessor):
         except NameError:
             self.logger.warning("Could not parse survey data for node: " + str(node))
 
-        # Store survey graph into datastream.
-        context.datastream.survey_topology = SurveyInfoStreamsData(node, vertices, edges)
+        latest_stored_graph = None
+        # Retrieve the latest stored datapoint, if it exists
+        streams = datastream.find_streams({'node': node.uuid})
+        if streams:
+            latest_stream = streams[len(streams) - 1]
+            datapoint_iterator = datastream.get_data(stream_id=latest_stream['stream_id'],
+                                                     granularity=latest_stream['highest_granularity'],
+                                                     start=latest_stream['latest_datapoint'])
+            try:
+                latest_stored_graph = datapoint_iterator[0]['v']
+            except IndexError:
+                pass
+
+        latest_graph = {
+            "v": vertices,
+            "e": edges
+        }
+
+        if latest_graph != latest_stored_graph:
+            # Store the latest graph into datastream.
+            context.datastream.survey_topology = SurveyInfoStreamsData(node, latest_graph)
         return context
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -93,6 +93,7 @@ class SurveyInfo(monitor_processors.NodeProcessor):
     def get_node_channel(self, context, node, frequency_band='2'):
         """
         Extracts the channel number from the context for the specified frequency band.
+
         :param context: Current context
         :param node: Node that is being processed
         :param frequency_band: either '2' or '5', corresponding to 2.4GHz and 5GHz.
@@ -109,6 +110,7 @@ class SurveyInfo(monitor_processors.NodeProcessor):
     def get_node_snr(self, context, node, frequency_band='2'):
         """
         Extracts the SNR from the context for the specified frequency band.
+
         :param context: Current context
         :param node: Node that is being processed
         :param frequency_band: either '2' or '5', corresponding to 2.4GHz and 5GHz.
@@ -134,6 +136,9 @@ class SurveyInfo(monitor_processors.NodeProcessor):
     def get_node_neighbors(self, context, node, frequency_band='2'):
         """
         Returns a dictionary of all access points in the vicinity along with the signal strength of each access point.
+
+        :param context: Current context
+        :param node: Node that is being processed
         :param: frequency_band: either '2' or '5', corresponding to 2.4GHz and 5GHz.
         :return: a dictionary of neighbors at the specified frequency band
         """
@@ -151,8 +156,8 @@ class SurveyInfo(monitor_processors.NodeProcessor):
         try:
             for radio in context.http.core.wireless.radios:
                 for neighbor in context.http.core.wireless.radios[str(radio)]['survey']:
-                    if neighbor['channel'] <= frequency_band_max_channel and neighbor['channel'] \
-                            >= frequency_band_min_channel:
+                    if neighbor['channel'] <= frequency_band_max_channel \
+                            and neighbor['channel'] >= frequency_band_min_channel:
                         return context.http.core.wireless.radios[radio]['survey']
         except KeyError:
             pass
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -1,0 +1,157 @@
+import datetime
+import pytz
+
+from django.utils.translation import gettext_noop
+
+from nodewatcher.core.monitor import models as monitor_models, processors as monitor_processors
+from nodewatcher.utils import ipaddr
+from nodewatcher.modules.monitor.sources.http import processors as http_processors
+
+DATASTREAM_SUPPORTED = False
+try:
+    from nodewatcher.modules.monitor.datastream import fields as ds_fields, models as ds_models
+    from nodewatcher.modules.monitor.datastream.pool import pool as ds_pool
+
+    class SurveyInfoStreams(ds_models.RegistryRootStreams):
+        snr = ds_fields.IntegerField(tags={
+            'title': gettext_noop("SNR"),
+            'description': gettext_noop("Node SNR."),
+            'visualization': {
+                'type': 'line',
+                'initial_set': True,
+                'time_downsamplers': ['mean'],
+                'value_downsamplers': ['min', 'mean', 'max'],
+                'minimum': 0.0,
+            },
+        })
+        neighbor_count = ds_fields.IntegerField(tags={
+            'title': gettext_noop("Neighbor count"),
+            'description': gettext_noop("Number of detected access points."),
+            'visualization': {
+                'type': 'line',
+                'initial_set': True,
+                'time_downsamplers': ['mean'],
+                'value_downsamplers': ['min', 'mean', 'max'],
+                'minimum': 0.0,
+            },
+        })
+
+        def get_module_name(self):
+            return 'monitor.http.survey'
+
+    class SurveyInfoStreamsData(object):
+        def __init__(self, node, frequency, channel, snr, neighbors):
+            self.node = node
+            self.frequency = frequency #either 2 or 5, corresponding to 2.4GHz and 5GHz
+            self.channel = channel
+            self.snr = snr
+            self.neighbors = neighbors
+            self.neighbor_count = len(self.neighbors)
+
+    ds_pool.register(SurveyInfoStreamsData, SurveyInfoStreams)
+
+    DATASTREAM_SUPPORTED = True
+except ImportError:
+    pass
+
+
+class SurveyInfo(monitor_processors.NodeProcessor):
+    """
+    Stores node's reported channel, SNR and neighbor data into the monitoring schema. Will
+    only run if HTTP monitor module has previously fetched data.
+    """
+
+    @monitor_processors.depends_on_context("http", http_processors.HTTPTelemetryContext)
+    def process(self, context, node):
+        """
+        Called for every processed node.
+
+        :param context: Current context
+        :param node: Node that is being processed
+        :return: A (possibly) modified context
+        """
+        # Should I go through the existing clients as well?
+        # existing_clients = {}
+        #for client in node.monitoring.network.clients():
+        #    existing_clients[client.client_id] = client
+
+        version = context.http.get_module_version("core.clients")
+        if version == 0:
+            # Unsupported version or data fetch failed (v0)
+            return context
+
+        print("incoming clients")
+        channel = self.getNodeChannel(context, node)
+        snr = self.getNodeSNR(context, node)
+        neighbors = self.getNodeNeighbors(context,node)
+        print(neighbors)
+        if DATASTREAM_SUPPORTED:
+            # Store client count into datastream.
+            context.datastream.monitor_http_clients = SurveyInfoStreamsData(node, "2", channel, snr, neighbors)
+
+        return context
+
+    def getNodeChannel(self, context, node, frequency_band="2"):
+        """
+        Extracts the channel number from the context for the specified frequency band.
+        :param context: Current context
+        :param node: Node that is being processed
+        :param frequency_band: either "2" or "5", corresponding to 2.4GHz and 5GHz.
+        :return: channel currently assigned to the node at the specified frequency band.
+        """
+        try:
+            for interface in context.http.core.wireless.interfaces:
+                if str(context.http.core.wireless.interfaces[interface].frequency)[:1] == frequency_band:
+                    return context.http.core.wireless.interfaces[interface].channel
+        except NameError:
+            print("Error parsing JSON file for " + node)
+
+    def getNodeSNR(self, context, node, frequency_band="2"):
+        """
+        Extracts the SNR from the context for the specified frequency band.
+        :param context: Current context
+        :param node: Node that is being processed
+        :param frequency_band: either "2" or "5", corresponding to 2.4GHz and 5GHz.
+        :return: SNR of the node at the specified frequency band.
+        """
+        try:
+            for interface in context.http.core.wireless.interfaces:
+                if str(context.http.core.wireless.interfaces[interface].frequency)[:1] == frequency_band:
+                    signal = context.http.core.wireless.interfaces[interface].signal
+                    noise = context.http.core.wireless.interfaces[interface].noise
+                    if (not signal or not noise):
+                        print("No SNR available for " + node)
+                        return 0
+                    else:
+                        return signal - noise
+        except NameError:
+            print("Error parsing JSON file for " + node)
+        except KeyError:
+            print("No SNR available for " + node)
+            return 0
+
+    def getNodeNeighbors(self, context, node, frequency_band="2"):
+        """
+        returns a dictionary of all access points in the vicinity along with the signal strength of each access point.
+        :param: frequency_band: either "2" or "5", corresponding to 2.4GHz and 5GHz.
+        :return: a dictionary of neighbors at the specified frequency band
+        """
+        if frequency_band == "2":
+            frequency_band_max_channel = 14
+            frequency_band_min_channel = 1
+        elif frequency_band == "5":
+            frequency_band_max_channel = 165
+            frequency_band_min_channel = 36
+        else:
+            print("Frequency band not recognized")
+            return
+
+        try:
+            for radio in context.http.core.wireless.radios:
+                for neighbor in context.http.core.wireless.radios[str(radio)]["survey"]:
+                    if neighbor["channel"] <= frequency_band_max_channel and neighbor["channel"] >= frequency_band_min_channel:
+                        return context.http.core.wireless.radios[radio]["survey"]
+        except KeyError:
+            pass
+        except NameError:
+            print("Error parsing JSON file for " + node)
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -42,7 +42,7 @@ try:
     class SurveyInfoStreamsData(object):
         def __init__(self, node, frequency, channel, snr, neighbors):
             self.node = node
-            self.frequency = frequency #either 2 or 5, corresponding to 2.4GHz and 5GHz
+            self.frequency = frequency  # either 2 or 5, corresponding to 2.4GHz and 5GHz
             self.channel = channel
             self.snr = snr
             self.neighbors = neighbors
@@ -83,7 +83,7 @@ class SurveyInfo(monitor_processors.NodeProcessor):
         print("incoming clients")
         channel = self.getNodeChannel(context, node)
         snr = self.getNodeSNR(context, node)
-        neighbors = self.getNodeNeighbors(context,node)
+        neighbors = self.getNodeNeighbors(context, node)
         print(neighbors)
         if DATASTREAM_SUPPORTED:
             # Store client count into datastream.
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -75,7 +75,7 @@ class SurveyInfo(monitor_processors.NodeProcessor):
         #for client in node.monitoring.network.clients():
         #    existing_clients[client.client_id] = client
 
-        version = context.http.get_module_version("core.wireless")
+        version = context.http.get_module_version('core.wireless')
         if version == 0:
             # Unsupported version or data fetch failed (v0)
             return context
@@ -85,16 +85,16 @@ class SurveyInfo(monitor_processors.NodeProcessor):
         neighbors = self.get_node_neighbors(context, node)
         if DATASTREAM_SUPPORTED:
             # Store client count into datastream.
-            context.datastream.monitor_http_clients = SurveyInfoStreamsData(node, "2", channel, snr, neighbors)
+            context.datastream.monitor_http_clients = SurveyInfoStreamsData(node, '2', channel, snr, neighbors)
 
         return context
 
-    def get_node_channel(self, context, node, frequency_band="2"):
+    def get_node_channel(self, context, node, frequency_band='2'):
         """
         Extracts the channel number from the context for the specified frequency band.
         :param context: Current context
         :param node: Node that is being processed
-        :param frequency_band: either "2" or "5", corresponding to 2.4GHz and 5GHz.
+        :param frequency_band: either '2' or '5', corresponding to 2.4GHz and 5GHz.
         :return: channel currently assigned to the node at the specified frequency band.
         """
         try:
@@ -104,12 +104,12 @@ class SurveyInfo(monitor_processors.NodeProcessor):
         except NameError:
             print("Error parsing JSON file for " + node)
 
-    def get_node_snr(self, context, node, frequency_band="2"):
+    def get_node_snr(self, context, node, frequency_band='2'):
         """
         Extracts the SNR from the context for the specified frequency band.
         :param context: Current context
         :param node: Node that is being processed
-        :param frequency_band: either "2" or "5", corresponding to 2.4GHz and 5GHz.
+        :param frequency_band: either '2' or '5', corresponding to 2.4GHz and 5GHz.
         :return: SNR of the node at the specified frequency band.
         """
         try:
@@ -128,16 +128,16 @@ class SurveyInfo(monitor_processors.NodeProcessor):
             print("No SNR available for " + node)
             return 0
 
-    def get_node_neighbors(self, context, node, frequency_band="2"):
+    def get_node_neighbors(self, context, node, frequency_band='2'):
         """
         returns a dictionary of all access points in the vicinity along with the signal strength of each access point.
-        :param: frequency_band: either "2" or "5", corresponding to 2.4GHz and 5GHz.
+        :param: frequency_band: either '2' or '5', corresponding to 2.4GHz and 5GHz.
         :return: a dictionary of neighbors at the specified frequency band
         """
-        if frequency_band == "2":
+        if frequency_band == '2':
             frequency_band_max_channel = 14
             frequency_band_min_channel = 1
-        elif frequency_band == "5":
+        elif frequency_band == '5':
             frequency_band_max_channel = 165
             frequency_band_min_channel = 36
         else:
@@ -146,9 +146,9 @@ class SurveyInfo(monitor_processors.NodeProcessor):
 
         try:
             for radio in context.http.core.wireless.radios:
-                for neighbor in context.http.core.wireless.radios[str(radio)]["survey"]:
-                    if neighbor["channel"] <= frequency_band_max_channel and neighbor["channel"] >= frequency_band_min_channel:
-                        return context.http.core.wireless.radios[radio]["survey"]
+                for neighbor in context.http.core.wireless.radios[str(radio)]['survey']:
+                    if neighbor['channel'] <= frequency_band_max_channel and neighbor['channel'] >= frequency_band_min_channel:
+                        return context.http.core.wireless.radios[radio]['survey']
         except KeyError:
             pass
         except NameError:
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -71,11 +71,6 @@ class SurveyInfo(monitor_processors.NodeProcessor):
         :return: A (possibly) modified context
         """
 
-        # Should I go through the existing clients as well?
-        # existing_clients = {}
-        #for client in node.monitoring.network.clients():
-        #    existing_clients[client.client_id] = client
-
         version = context.http.get_module_version('core.wireless')
         if version == 0:
             # Unsupported version or data fetch failed (v0)
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -93,7 +93,7 @@ class SurveyInfo(monitor_processors.NodeProcessor):
                 stream_id=streams[0]['stream_id'],
                 granularity=streams[0]['highest_granularity'],
                 start=streams[0]['latest_datapoint'],
-                reverse=True
+                reverse=True,
             )
             try:
                 latest_stored_graph = datapoint_iterator[0]['v']
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -86,8 +86,9 @@ class SurveyInfo(monitor_processors.NodeProcessor):
 
         latest_stored_graph = None
         # Retrieve the latest stored datapoint, if it exists
-        streams = datastream.find_streams({'node': node.uuid})
+        streams = datastream.find_streams({'node': node.uuid, 'module': 'monitor.http.survey'})
         if streams:
+            assert len(streams) == 1
             latest_stream = streams[len(streams) - 1]
             datapoint_iterator = datastream.get_data(
                 stream_id=latest_stream['stream_id'],
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -1,10 +1,6 @@
-import datetime
-import pytz
-
 from django.utils.translation import gettext_noop
 
-from nodewatcher.core.monitor import models as monitor_models, processors as monitor_processors
-from nodewatcher.utils import ipaddr
+from nodewatcher.core.monitor import processors as monitor_processors
 from nodewatcher.modules.monitor.sources.http import processors as http_processors
 
 from django_datastream import datastream
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -10,14 +10,19 @@ from nodewatcher.modules.monitor.datastream.pool import pool as ds_pool
 
 class SurveyInfoStreams(ds_models.RegistryRootStreams):
     neighbor_graph = ds_fields.GraphField(tags={
-        'title': _("Neighbor topology"),
-        'description': _("Neighbor topology."),
+        'title': _("Neighborhood topology"),
+        'description': _("Neighborhood topology."),
     })
 
     def get_module_name(self):
         return 'monitor.http.survey'
 
     def get_stream_highest_granularity(self):
+        # The nodewatcher-agent performs a survey once every ~240 monitoring intervals according to
+        # https://github.com/wlanslovenija/nodewatcher-agent/blob/master/modules/wireless.c#L362
+        # and a monitoring run is performed every 30 seconds according to
+        # https://github.com/wlanslovenija/nodewatcher-agent/blob/73f2b25db2c34ae7d70904bac31379a2243bade0/modules/wireless.c#L437.
+        # So a survey is performed once every two hours, meaning that we use hourly granularity.
         return datastream.Granularity.Hours
 
 
@@ -35,10 +40,11 @@ ds_pool.register(SurveyInfoStreamsData, SurveyInfoStreams)
 
 class SurveyInfo(monitor_processors.NodeProcessor):
     """
-    Stores neighbor's reported channel, BSSID, SSID and signal strength into the datastream. A star shaped graph
-    is constructed. The source vertex contains an array of all BSSIDs of that node. Source vertex is stored according
-    to its UUID, wherease all neighbors are stored according to their BSSIDs. Will only run if HTTP monitor module
-    has previously fetched data.
+    Accesses the data collected during a WiFi survey of neighboring wireless access points / routers. Stores channel,
+    BSSID, SSID and signal strength for each neighbor into the datastream. A star shaped graph is constructed. The
+    source vertex contains an array of all BSSIDs of that node. Source vertex is stored according to its UUID,
+    wherease all neighbors are stored according to their BSSIDs. Will only run if HTTP monitor module has previously
+    fetched data.
     """
 
     @monitor_processors.depends_on_context("http", http_processors.HTTPTelemetryContext)
@@ -55,17 +61,19 @@ class SurveyInfo(monitor_processors.NodeProcessor):
         if version == 0:
             # Unsupported version or data fetch failed
             return context
+
         edges, vertices = [], []
         try:
-            vertex_bssids = []
-            for interface in context.http.core.wireless.interfaces.values():
-                vertex_bssids.append(interface['bssid'])
-            vertex = dict(i=str(node.uuid), bssids=vertex_bssids)
-            vertices.append(vertex)
+            source_vertex_bssids = [interface.bssid for interface in context.http.core.wireless.interfaces.values()]
+            source_vertex = {
+                'i': str(node.uuid),
+                'bssids': source_vertex_bssids,
+            }
+            vertices.append(source_vertex)
             for radio in context.http.core.wireless.radios.values():
                 for neighbor in radio.survey:
-                    vertices.append(dict(i=neighbor['bssid']))
-                    # Add an edge for that neighbor
+                    vertices.append({'i': neighbor['bssid']})
+                    # Add an edge for that neighbor.
                     edge = {'f': str(node.uuid), 't': neighbor['bssid']}
                     edge['channel'] = neighbor['channel']
                     edge['signal'] = neighbor['signal']
@@ -74,7 +82,8 @@ class SurveyInfo(monitor_processors.NodeProcessor):
         except KeyError:
             pass
         except NameError:
-            self.logger.warning("Error parsing JSON file for " + str(node))
+            self.logger.warning("Could not parse survey data for node: " + str(node))
+
         # Store survey graph into datastream.
         context.datastream.survey_topology = SurveyInfoStreamsData(node, vertices, edges)
         return context
--- a/nodewatcher/modules/monitor/http/survey/processors.py
+++ b/nodewatcher/modules/monitor/http/survey/processors.py
@@ -10,7 +10,7 @@ try:
     from nodewatcher.modules.monitor.datastream import fields as ds_fields, models as ds_models, base as ds_base
     from nodewatcher.modules.monitor.datastream.pool import pool as ds_pool
 
-    class SurveyInfoStreams(ds_base.StreamsBase):
+    class SurveyInfoStreams(ds_models.RegistryRootStreams):
         neighbor_graph = ds_fields.GraphField(tags={
             'title': gettext_noop("Neighbor topology"),
             'description': gettext_noop("Neighbor topology."),
@@ -19,17 +19,9 @@ try:
         def get_module_name(self):
             return 'monitor.http.survey'
 
-        def get_stream_query_tags(self):
-            return {'module': 'survey'}
-
-        def get_stream_tags(self):
-            return {'module': 'survey'}
-
-        def get_stream_highest_granularity(self):
-            return datastream.Granularity.Minutes
-
     class SurveyInfoStreamsData(object):
-        def __init__(self, vertices, edges):
+        def __init__(self, node, vertices, edges):
+            self.node = node
             self.neighbor_graph = {
                 'v': vertices,
                 'e': edges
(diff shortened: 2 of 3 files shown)


nodewatcher/settings.py
```diff
@@ -266,6 +266,7 @@ def user_url(user):
     'nodewatcher.modules.monitor.http.resources',
     'nodewatcher.modules.monitor.http.interfaces',
     'nodewatcher.modules.monitor.http.clients',
+    'nodewatcher.modules.monitor.http.survey',
```
CdavM: Do I have to add this into MONITOR_RUNS processors list as well?

nodewatcher/settings.py
```diff
@@ -452,6 +452,7 @@ def user_url(user):
     'nodewatcher.modules.monitor.http.resources.processors.SystemStatus',
     'nodewatcher.modules.monitor.http.interfaces.processors.DatastreamInterfaces',
     'nodewatcher.modules.monitor.http.clients.processors.ClientInfo',
+    'nodewatcher.modules.monitor.http.survey.processors.SurveyInfo',
```
CdavM, line 455: Do I have to add this line into MONITOR_RUNS processors list as well?

nodewatcher/modules/monitor/http/survey/cgm.py
```diff
@@ -0,0 +1,10 @@
+from nodewatcher.core.generator.cgm import base as cgm_base
+
+
+@cgm_base.register_platform_module('openwrt', 50)
+def clients(node, cfg):
+    """
+    Register packages needed for reporting the clients.
+    """
+
+    cfg.packages.update(['nodewatcher-agent-mod-clients'])
```
CdavM: Is the 'clients' method necessary for my module? I copied it from clients module but I don't think I'm using it.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,157 @@
...
+        def get_module_name(self):
+            return 'monitor.http.survey'
+
+    class SurveyInfoStreamsData(object):
+        def __init__(self, node, frequency, channel, snr, neighbors):
+            self.node = node
+            self.frequency = frequency #either 2 or 5, corresponding to 2.4GHz and 5GHz
+            self.channel = channel
+            self.snr = snr
+            self.neighbors = neighbors
```
CdavM: I'm saving the dictionary of neighbors here, but I'm only plotting the number of neighbors onto the plot. I think this is the right way of doing it, but @mitar mentioned he'd like to construct a graph and save that instead.

nodewatcher/modules/monitor/http/survey/cgm.py
kostko: CGMs (configuration generating modules) are functions, which generate firmware configuration. In this instance, you just specify that the `nodewatcher-agent-mod-clients` OpenWrt package should be installed for each node. However, I think that you use output from the `wireless` module and not `clients`. So you should change this to depend on the correct module.

The modules are those of [nodewatcher-agent](https://github.com/wlanslovenija/nodewatcher-agent/tree/master/modules), with packages defined in [our OpenWrt package feed](https://github.com/wlanslovenija/firmware-packages-opkg/blob/master/util/nodewatcher-agent/Makefile#L87).

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,156 @@
...
+        :param context: Current context
+        :param node: Node that is being processed
+        :return: A (possibly) modified context
+        """
+        # Should I go through the existing clients as well?
+        # existing_clients = {}
+        #for client in node.monitoring.network.clients():
+        #    existing_clients[client.client_id] = client
+
+        version = context.http.get_module_version("core.clients")
```
kostko: I think that you are using output from the `core.wireless` module, so you should check its version and not `core.clients`.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,156 @@
...
+        snr = self.getNodeSNR(context, node)
+        neighbors = self.getNodeNeighbors(context, node)
+        print(neighbors)
+        if DATASTREAM_SUPPORTED:
+            # Store client count into datastream.
+            context.datastream.monitor_http_clients = SurveyInfoStreamsData(node, "2", channel, snr, neighbors)
+
+        return context
+
+    def getNodeChannel(self, context, node, frequency_band="2"):
```
kostko: Do not use camel case for method or variable names, you should always use underscores. Also elswhere.

kostko: Also you should use single quotes for internal constants and double quotes for user-facing strings. So it should be `'2'`, also elsewhere.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        Extracts the channel number from the context for the specified frequency band.
+        :param context: Current context
+        :param node: Node that is being processed
+        :param frequency_band: either "2" or "5", corresponding to 2.4GHz and 5GHz.
+        :return: channel currently assigned to the node at the specified frequency band.
+        """
+        try:
```
kostko: Empty line between documentation blocks and function body.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,156 @@
...
+        print(neighbors)
+        if DATASTREAM_SUPPORTED:
+            # Store client count into datastream.
+            context.datastream.monitor_http_clients = SurveyInfoStreamsData(node, "2", channel, snr, neighbors)
+
+        return context
+
+    def getNodeChannel(self, context, node, frequency_band="2"):
+        """
+        Extracts the channel number from the context for the specified frequency band.
```
kostko: Empty line between description and arguments.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,156 @@
...
+        :param node: Node that is being processed
+        :param frequency_band: either "2" or "5", corresponding to 2.4GHz and 5GHz.
+        :return: channel currently assigned to the node at the specified frequency band.
+        """
+        try:
+            for interface in context.http.core.wireless.interfaces:
+                if str(context.http.core.wireless.interfaces[interface].frequency)[:1] == frequency_band:
+                    return context.http.core.wireless.interfaces[interface].channel
+        except NameError:
+            print("Error parsing JSON file for " + node)
```
kostko: Do not use `print` for logging, use `self.logger` instead, eg. `self.logger.warning("Error parsing JSON ...")`. Also elsewhere.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,156 @@
...
+                        return signal - noise
+        except NameError:
+            print("Error parsing JSON file for " + node)
+        except KeyError:
+            print("No SNR available for " + node)
+            return 0
+
+    def getNodeNeighbors(self, context, node, frequency_band="2"):
+        """
+        returns a dictionary of all access points in the vicinity along with the signal strength of each access point.
```
kostko: Break long lines, start documentation blocks with a capital letter.

nodewatcher/settings.py
kostko, line 455: It is already included in `MONITOR_RUNS`, check the `telemetry` and `telemetry-push` run definitions, which reference `TELEMETRY_PROCESSOR_PIPELINE`.

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: Addressed with e0dbee7

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: CamelCase fixed with 8d885ac
Quote conventions fixed with 6edf9d6

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: Fixed with ae0e0c8

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: fixed with  c1c0868

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: Documentation blocks fixed with 6834bb8

What length is considered long? My IDE wraps words at 120 characters.

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: Fixed with e304340

nodewatcher/modules/monitor/http/survey/processors.py
mitar: I mostly wrap so that it is similar to the code around it. But max 120 is good for both code and comments. But for code sometimes it is simply more readable if it is longer. So that is also OK. It is really a question how to make it the most readable.

nodewatcher/modules/monitor/http/survey/cgm.py
```diff
@@ -0,0 +1,10 @@
+from nodewatcher.core.generator.cgm import base as cgm_base
+
+
+@cgm_base.register_platform_module('openwrt', 50)
+def clients(node, cfg):
```
mitar: Name of the function is misleading?

nodewatcher/modules/monitor/http/survey/cgm.py
```diff
@@ -0,0 +1,10 @@
...
+    """
+    Register packages needed for reporting the clients.
```
mitar: Comment is misleading.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,165 @@
...
+    def process(self, context, node):
+        """
+        Called for every processed node.
+
+        :param context: Current context
+        :param node: Node that is being processed
+        :return: A (possibly) modified context
+        """
+
+        # Should I go through the existing clients as well?
```
mitar: Which clients???

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,165 @@
...
+            },
+        })
+
+        def get_module_name(self):
+            return 'monitor.http.survey'
+
+    class SurveyInfoStreamsData(object):
+        def __init__(self, node, frequency, channel, snr, neighbors):
+            self.node = node
+            self.frequency = frequency  # either 2 or 5, corresponding to 2.4GHz and 5GHz
```
mitar: Why is this needed? Just store channel.

mitar:
Please use full sentences for commit messages. (Dot at the end.)

mitar:
So currently you are not storing survey data into the database. You probably want more than just count of neighbors to compute your stuff, no?

kostko:
An example of how to store a graph into datastream, see:
- https://github.com/wlanslovenija/nodewatcher/blob/development/nodewatcher/modules/monitor/topology/processors.py

mitar:
For example how to run periodic tasks, see:
- https://github.com/wlanslovenija/nodewatcher/blob/development/nodewatcher/core/generator/tasks.py

So then daily run analysis (some code under `nodewatcher.modules.analysis.`) and then store data into your module's model (a model with foreign key to WiFi interface config model) where you store suggested value.

Example how you define a warning:
- https://github.com/wlanslovenija/nodewatcher/blob/development/nodewatcher/modules/monitor/validation/interfaces/events.py

Example how you create a warning:
- https://github.com/wlanslovenija/nodewatcher/blob/development/nodewatcher/modules/monitor/validation/interfaces/processors.py

(In your case a background task will create a warning if a node uses not-suggested channel.)

nodewatcher/modules/monitor/http/survey/cgm.py
CdavM: Addressed with b80cdc1 and 69d29ee

nodewatcher/modules/monitor/http/survey/cgm.py
CdavM: Addressed with 3d69be5

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: So which channel do I store? The 2.4ghz channel or 5ghz channel?

mitar: Look how we store channel and channel width for WiFi interface in monitoring schema.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,160 @@
...
+            self.channel = channel
+            self.snr = snr
+            self.neighbors = neighbors
```
CdavM: Doesn't this line store the neighbors JSON object? The neighbor_count variable is only there for plotting the number of neighbors.

mitar: Find me this object stored. Or, show me how you will read it from the database.

CdavM:
I am now creating a graph. I can see it in the tokumx database, as well as in the datastream (via /api/v1/stream) but I cannot see a visualization of the graph. I was expecting the graph under all the plots on the page of that client. Is this a bug or a feature? Or possibly a problem with my implementation?

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -111,17 +109,24 @@ def get_node_snr(self, context, node, frequency_band='2'):
...
-                    signal = context.http.core.wireless.interfaces[interface].signal
-                    noise = context.http.core.wireless.interfaces[interface].noise
-                    if (not signal or not noise):
-                        self.logger.warning("No SNR available for " + str(node))
-                        return 0
-                    else:
-                        return signal - noise
+                # create an edge for every interface
+                edge = {}
+                edge = {'f': str(node), 't': destination_id}
```
CdavM: I'm currently using the name of the node for the 'from' field but I am using BSSIDs for the 'to' field. What do you think? Should I use BSSIDs or MAC addresses or something else for the source node? The benefit of the current implementation is that we only have one source node, whereas BSSIDs differ for different interfaces.

mitar:
> I was expecting the graph under all the plots on the page of that client. Is this a bug or a feature?

Now, we do not support rendering of graph types on the node page. So do not select it for visualization. You are just storing data. And this is it. So that you can use it later.

If you want, you can implement visualization for this data type. But after you finish everything else.

So you are saying that this pull request is finished? It is storing survey data correctly?

Have you verified if noise is available on nodes for survey data?

mitar:
It seems [noise is not available](http://lxr.mein.io/source/iwinfo/include/iwinfo.h#L140). So let's then use only signal.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,92 @@
+import datetime
+import pytz
+
+from django.utils.translation import gettext_noop
```
mitar: Always import this as `_`.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,92 @@
...
+
+from nodewatcher.core.monitor import models as monitor_models, processors as monitor_processors
+from nodewatcher.utils import ipaddr
```
mitar: Where are you using this import?

mitar: Cleanup imports.

nodewatcher/modules/monitor/http/survey/processors.py
mitar: (Feel free to fix it also in the topology processor.)

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,92 @@
...
+    from nodewatcher.modules.monitor.datastream import fields as ds_fields, models as ds_models, base as ds_base
+    from nodewatcher.modules.monitor.datastream.pool import pool as ds_pool
+
+    class SurveyInfoStreams(ds_base.StreamsBase):
+        neighbor_graph = ds_fields.GraphField(tags={
+            'title': gettext_noop("Neighbor topology"),
+            'description': gettext_noop("Neighbor topology."),
+            'visualization': {
+                'type': 'graph',
+                'initial_set': True,
```
mitar: I think this does not have to be specified. This means that we want to display this on the node page with initial graphs. I do not think this is reasonable. (And we do not support it anyway.)

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,92 @@
...
+            return 'monitor.http.survey'
+
+        def get_stream_query_tags(self):
+            return {'module': 'survey'}
+
+        def get_stream_tags(self):
+            return {'module': 'survey'}
+
+        def get_stream_highest_granularity(self):
+            return datastream.Granularity.Minutes
```
mitar: You should check how often are we doing survey on nodes. Because we are definitely not doing it every minute. (Because it takes radio offline.) So maybe we are doing it every 30 minutes, or even every hour. Check and then configure this accordingly.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,92 @@
...
+                'type': 'graph',
+                'initial_set': True,
+            }
+        })
+
+        def get_module_name(self):
+            return 'monitor.http.survey'
+
+        def get_stream_query_tags(self):
+            return {'module': 'survey'}
```
mitar: So the network topology data is per whole network. But these graphs are per node. So you should store node information into this tags. See how it is done in `RegistryRootStreams` as an example. And `StreamsBase` for what you have to define (you do not have to define `get_module_name`, because this is in  `RegistryRootStreams`, which you are not using here). You might simply use `RegistryRootStreams` here as base and just define `get_module_name`?

mitar: @kostko?

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,92 @@
...
+    ds_pool.register(SurveyInfoStreamsData, SurveyInfoStreams)
+
+    DATASTREAM_SUPPORTED = True
+except ImportError:
+    pass
+
+
+class SurveyInfo(monitor_processors.NodeProcessor):
+    """
+    Stores neighbor's reported channel, BSSID and signal strength into the monitoring schema. A star shaped graph
```
mitar: Are you really storing anything into the monitoring schema? I thin you are storing just into datastream?

In this case you might want to get rid of whole `DATASTREAM_SUPPORTED` stuff and simply require datastream for this module. Like topology requires it. Because the point of `DATASTREAM_SUPPORTED` is that you can store something only into monitoring schema, but not into time-series database. So that you have only latest values.

But you are not storing latest values, because monitoring schema does not support graph types anyway.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,92 @@
...
+        version = context.http.get_module_version('core.wireless')
+        if version == 0:
+            # Unsupported version or data fetch failed
+            return context
+        edges, vertices = [], []
+        try:
+            vertices.append(dict(i=str(node)))
+            for radio in context.http.core.wireless.radios:
+                for neighbor in context.http.core.wireless.radios[str(radio)]['survey']:
+                    # add an edge for that neighbor
```
mitar: Start with an upper-case letter. We are also moving to style where all comments should be full sentences.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,92 @@
...
+        :return: A (possibly) modified context
+        """
+
+        version = context.http.get_module_version('core.wireless')
+        if version == 0:
+            # Unsupported version or data fetch failed
+            return context
+        edges, vertices = [], []
+        try:
+            vertices.append(dict(i=str(node)))
```
mitar: I do not think that `str(node)` is the best way. This is human-oriented representation of the node, not what you want to store into the database. Probably you want `node.uuid`?

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,92 @@
...
+            for radio in context.http.core.wireless.radios:
```
mitar: You might want to use `ontext.http.core.wireless.radios.items()` to iterate over both key and value?

mitar:
For graph data. I would be storing node uuid as `i`, and `from` should then be this `uuid`. For `to`, I would store `bssid`. But into vertex object of `from` I would also store all local `bssid`s a node has (you can find them inside interfaces). So an array of bssids. This can then allow one to take only survey data and compute the whole graph. And has uuid of our nodes as well.

nodewatcher/modules/monitor/http/survey/processors.py
mitar: I would also store SSID for vertices. So that it will be easier to visualize networks around a node with names. (Only for `to` node, so `to` node has SSID, while `from` node has list of BSSIDs.)

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: So why would we need the key? Would we also store the key into an edge? So would an edge look like:
edge = {'f': node.uuid, 't': bssid, 'channel': 1, 'signal': 2, 'radio': 'phy1'}
?

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: We're only storing values into the datastream.

I looked at the topology module and I don't see how to require access to the datastream. It looks like topology just assumes that it can access the datastream?

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: Addressed with ae1e908.

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: addressed with 476e48b.

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: I'm no longer inheriting from `ds_base.StreamsBase` but rather from `RegistryRootStreams`. Implemented in 19edfd8

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: Added support for SSIDs in the edges with 9757ad5

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: I am now using `.values()` to iterate over the values.

CdavM:
We're now storing source node uuid as `i`, and `f` in an edge is this uuid. For `t`, the bssid is used. Vertex object stores an array of bssids a node has (from interfaces).

nodewatcher/modules/monitor/http/survey/processors.py
mitar: > So why would we need the key?

I think you got confused here. :-) I was not saying about storing anything.

nodewatcher/modules/monitor/http/survey/processors.py
mitar: > I looked at the topology module and I don't see how to require access to the datastream. It looks like topology just assumes that it can access the datastream?

Yes, you can do that. :-) Just assume you have datastream.

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: How do I find this? I checked `iw` and `iwinfo` in the ap and I found a way to perform a scan (`iw dev wlan1 scan`) but I don't know where these settings would be stored. I went through all the options `iw` lists and did not find scan periodicity settings

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
+from django.utils.translation import gettext_noop
```
mitar: Still not using `_`.

nodewatcher/modules/monitor/http/survey/processors.py
mitar: This is done by [nodewatcher agent](https://github.com/wlanslovenija/nodewatcher-agent). See here: https://github.com/wlanslovenija/nodewatcher-agent/blob/master/modules/wireless.c#L362

mitar: And: https://github.com/wlanslovenija/nodewatcher-agent/blob/73f2b25db2c34ae7d70904bac31379a2243bade0/modules/wireless.c#L437

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: Ah, okay. I see what you meant. 0b118ac

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: So according to https://github.com/wlanslovenija/nodewatcher-agent/blob/73f2b25db2c34ae7d70904bac31379a2243bade0/modules/wireless.c#L437 a node runs a monitoring cycle every 30 seconds (is seconds the correct time unit?) and a scan is performed once every 240 monitoring cycles so that's about once every two hours? So should we set the granularity to hourly or 6hours?

mitar: It should be smaller. So one hour.

But then also this survey processor should be storing data only once an hour. So you have two options: it runs for every monitor run and most of them are dummy and do not do anything, or it is required to be run in its own run.

I prefer the first option.

CdavM: OK, I changed the granularity option to Hours: aab02ad

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -2,9 +2,6 @@
 
 from nodewatcher.core.monitor import processors as monitor_processors
 from nodewatcher.modules.monitor.sources.http import processors as http_processors
-
-from django_datastream import datastream
```
mitar: Not sure if this is the change we were talking. Why this? Topology processor has it. You should be removing that try except statement.

mitar:
So finish the things and tell me when you think the pull request is ready for next code review.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,77 @@
+from django.utils.translation import gettext_noop as _
+from django_datastream import datastream
```
mitar, line 6: Empty line between groups of imports.

nodewatcher/modules/monitor/http/survey/processors.py
mitar: Update docstring to match code.

CdavM:
I can't find the comment about importing datastream anymore, but I removed that import because it wasn't used anywhere but after changing the granularity to Hours I imported datastream again. I'm done with all the changes requested, can you take another look?

mitar:
I removed that comment because I noticed that it was on the old version of the code.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+
+from nodewatcher.core.monitor import processors as monitor_processors
+from nodewatcher.modules.monitor.sources.http import processors as http_processors
+from nodewatcher.modules.monitor.datastream import fields as ds_fields, models as ds_models
+from nodewatcher.modules.monitor.datastream.pool import pool as ds_pool
+
+
+class SurveyInfoStreams(ds_models.RegistryRootStreams):
+    neighbor_graph = ds_fields.GraphField(tags={
+        'title': _("Neighbor topology"),
```
mitar: Nitpicking: Neighbor -> Neighborhood?

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+        'description': _("Neighbor topology."),
+    })
+
+    def get_module_name(self):
+        return 'monitor.http.survey'
+
+    def get_stream_highest_granularity(self):
+        return datastream.Granularity.Hours
```
mitar, line 28: Add a comment how you computed this. And when should this be changed.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+            'e': edges
+        }
+
+
+ds_pool.register(SurveyInfoStreamsData, SurveyInfoStreams)
+
+
+class SurveyInfo(monitor_processors.NodeProcessor):
+    """
+    Stores neighbor's reported channel, BSSID, SSID and signal strength into the datastream. A star shaped graph
```
mitar: Not just one neighbor's, no? But all.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+
+        version = context.http.get_module_version('core.wireless')
+        if version == 0:
+            # Unsupported version or data fetch failed
+            return context
+        edges, vertices = [], []
+        try:
+            vertex_bssids = []
+            for interface in context.http.core.wireless.interfaces.values():
+                vertex_bssids.append(interface['bssid'])
```
mitar: interface.bssid`, no?

mitar: BTW, this things could be made as an array comprehension:

```
vertex_bssids = (interface.bssid for interface in context.http.core.wireless.interfaces.values())
```

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+            vertex = dict(i=str(node.uuid), bssids=vertex_bssids)
```
mitar:
```
vertex = {
  'i': str(node.uuid),
  'bssids': vertex_bssids,
}
```

Maybe nicer?

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+            vertices.append(vertex)
+            for radio in context.http.core.wireless.radios.values():
+                for neighbor in radio.survey:
+                    vertices.append(dict(i=neighbor['bssid']))
```
mitar:
```
vertices.append({'i': neighbor.bssid}))
```

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+                    # Add an edge for that neighbor
```
mitar: Dot at the end.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+                    edge = {'f': str(node.uuid), 't': neighbor['bssid']}
```
mitar, line 76: `neighbor.bssid`

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+                    edge['channel'] = neighbor['channel']
```
mitar: `neighbor.channel`, no?

mitar:
Just style comments. Otherwise looks good. You tested it and it works for you? It stored data and you can access data through HTTP API, yes?

@kostko, can you look as well?

mitar:
So after you update small style changes, just go and create a new pull request from new branch (based on this branch) where you start adding data analytics. And start working on that.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+                    edge['signal'] = neighbor['signal']
+                    edge['ssid'] = neighbor['ssid']
+                    edges.append(edge)
+        except KeyError:
+            pass
+        except NameError:
+            self.logger.warning("Error parsing JSON file for " + str(node))
+        # Store survey graph into datastream.
```
kostko: Empty line above this one.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+        :param context: Current context
+        :param node: Node that is being processed
+        :return: A (possibly) modified context
+        """
+
+        version = context.http.get_module_version('core.wireless')
+        if version == 0:
+            # Unsupported version or data fetch failed
+            return context
+        edges, vertices = [], []
```
kostko, line 64: Empty line above this one.

nodewatcher/modules/monitor/http/survey/processors.py
kostko, line 28: Hm, the `telemetry` monitoring run (where you added the survey module to in settings) fetches data every few minutes. Is this the correct granularity?

mitar, line 28: Ah, you are right. I told him to make it so that it is run every few minutes, but that that most runs are dummy and are stored only every hour (or maybe even every time data is different from the last time).

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,80 @@
...
+
+from django_datastream import datastream
+
+from nodewatcher.core.monitor import processors as monitor_processors
+from nodewatcher.modules.monitor.sources.http import processors as http_processors
+from nodewatcher.modules.monitor.datastream import fields as ds_fields, models as ds_models
+from nodewatcher.modules.monitor.datastream.pool import pool as ds_pool
+
+
+class SurveyInfoStreams(ds_models.RegistryRootStreams):
```
mitar, line 14: @kostko: Is use of `RegistryRootStreams` here OK?

kostko:
Data collection part seems good. Is there a UI planned for visualizing the collected data?

nodewatcher/modules/monitor/http/survey/processors.py
kostko, line 14: Yes.

mitar:
> Data collection part seems good. Is there a UI planned for visualizing the collected data?

Not as part of GSoC. I might do something for that through dashboard work.

This will be used for data analysis modules.

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: I tried that and for some reason I'm getting an 
`AttributeError: 'dict' object has no attribute 'channel'`
I thought that you can access keys as object fields by default.

nodewatcher/modules/monitor/http/survey/processors.py
CdavM, line 28: Hm, so should we change it back to minute granularity?

nodewatcher/modules/monitor/http/survey/processors.py
mitar: Oh, `survey` is an array, so the returned things are simple dicts and not our context objects. @kostko, you agree?

Then leave it as it is. But be careful, if there is no `channel` in the `neighbor` dict, for example, this will throw an exception `KeyError`. You could maybe do:

```
if 'channel' in neighbor:
  edge['channel'] = neighbor['channel']
```

Or you can skip the whole edge, if you want. This is on you to decide how you want to deal with invalid/incomplete data. Currently, `KeyError` is then catched and the whole parsing stops, and only data parsed until then is stored. This is a bit strange. I would say we should or continue parsing other iterations, or do not store anything. Not partial data. @kostko?

mitar: BTW, I am not saying you should do `if 'channel' in neighbor:`, I am just showing you to think about that issue. And we already have some code to think about that (try/except) so you do not have to necessary do anything. But we should still all discuss what should be the best approach and we can improve it later.

nodewatcher/modules/monitor/http/survey/processors.py
mitar, line 28: NOOO. Can you please read the whole discussion about this. I am not sure what you are not getting. Think about it please. So how often does node do survey? How often do we read data from the node? And how often do we want then to store data into the node? And how often does the processor run?

There is little use of storing same data every minute for an hour, if node is changing the data it reports only every hour, no?

CdavM, line 28: Yes, but your comment "Ah, you are right" was confusing because you expressed agreement with Kostko who was doubting whether this was the correct granularity?

Anyway, I didn't change it and added comments about computing it.

CdavM:
I tested it and can see the data through the HTTP API stream.

nodewatcher/modules/monitor/http/survey/processors.py
mitar, line 28: Yes, he is right that with current code this is wrong. You have to fix it.

The important part for you is:

> I told him to make it so that it is run every few minutes, but that that most runs are dummy and are stored only every hour (or maybe even every time data is different from the last time).

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: After thinking about it, I created a tuple of the fields we're interested in and we then check if they're present before adding them. d799f03

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,105 @@
...
+    })
+
+    def get_module_name(self):
+        return 'monitor.http.survey'
+
+    def get_stream_highest_granularity(self):
+        # The nodewatcher-agent performs a survey once every ~240 monitoring intervals according to
+        # https://github.com/wlanslovenija/nodewatcher-agent/blob/master/modules/wireless.c#L362
+        # and a monitoring run is performed every 30 seconds according to
+        # https://github.com/wlanslovenija/nodewatcher-agent/blob/73f2b25db2c34ae7d70904bac31379a2243bade0/modules/wireless.c#L437.
```
mitar: Make link against master branch as well.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,105 @@
...
+        try:
+            source_vertex_bssids = [interface.bssid for interface in context.http.core.wireless.interfaces.values()]
+            source_vertex = {
+                'i': str(node.uuid),
+                'bssids': source_vertex_bssids,
+            }
+            vertices.append(source_vertex)
+            for radio in context.http.core.wireless.radios.values():
+                for neighbor in radio.survey:
+                    vertices.append({'i': neighbor['bssid']})
```
mitar, line 74: What if the same BSSID is present multiple times. Is this possible? Virtual SSIDs?

CdavM, line 74: The way I understand it, a BSSID is supposed to be unique (unlike a (virtual) SSID that multiple BSSIDs can share). But you identify a particular WAP through its BSSID, just like you identify a computer on the network by its MAC. So while you can spoof the BSSID (or MAC address) I don't know if you want to build in some protection against that? Do you want me to check if one exists before adding it?

CdavM, line 74: https://en.wikipedia.org/wiki/Service_set_(802.11_network)#Basic_service_set_identification_.28BSSID.29

mitar, line 74: But if the network is on 2.4 GHz and 5 GHz, does it have to have different BSSID?

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: Added with 848486b

nodewatcher/modules/monitor/http/survey/processors.py
CdavM, line 74: My understanding is that you need two antennas to broadcast on two different frequencies. And each antenna has a BSSID associated with it. Otherwise your computer couldn't distinguish between connecting to the 2.4 or 5GHz band? For instance, look at the interfaces at https://github.com/CdavM/swoon/blob/master/test_data/10.20.32.40.feed 
wlan0 (5GHz) has a bssid of 04:18:D6:21:60:44 and wlan1 (2.4GHz) has a bssid of 04:18:D6:22:60:44

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,106 @@
...
+                    edges.append(edge)
+        except KeyError:
+            pass
+        except NameError:
+            self.logger.warning("Could not parse survey data for node: " + str(node))
+
+        latest_stored_graph = None
+        # Retrieve the latest stored datapoint, if it exists
+        streams = datastream.find_streams({'node': node.uuid})
+        if streams:
```
mitar, line 89: `streams` is an object and I think it will always be true, not sure though.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,106 @@
...
+            latest_stream = streams[len(streams) - 1]
```
mitar: `streams[-1]` works in Python.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,106 @@
...
+            datapoint_iterator = datastream.get_data(stream_id=latest_stream['stream_id'],
```
mitar: This is not the style we use. We use:

```
datapoint_iterator = datastream.get_data(
    stream_id=latest_stream['stream_id'],
    granularity=latest_stream['highest_granularity'],
    start=latest_stream['latest_datapoint'],
)
```

nodewatcher/modules/monitor/http/survey/processors.py
CdavM, line 74: So this means we only have one edge between nodes in our graph but this should be fine.

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: But not for Mongo cursors:
`IndexError: Cursor instances do not support negativeindices`

nodewatcher/modules/monitor/http/survey/processors.py
mitar, line 74: OK. Sounds good.

nodewatcher/modules/monitor/http/survey/processors.py
mitar: Wouldn't it be better to simply do `reverse=True` and take the latest entry? Because otherwise there might be some race condition here between you reading `latest_stream` and asking for data.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,106 @@
...
+                                                     granularity=latest_stream['highest_granularity'],
+                                                     start=latest_stream['latest_datapoint'])
+            try:
+                latest_stored_graph = datapoint_iterator[0]['v']
+            except IndexError:
+                pass
+
+        latest_graph = {
+            "v": vertices,
```
mitar: Use `'` for constant strings (not-user facing strings).

nodewatcher/modules/monitor/http/survey/processors.py
CdavM, line 89: But can it return an empty iterator? So should I rather check 
`if len(streams) > 0:`
 ?

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,106 @@
...
+            "e": edges
+        }
+
+        if latest_graph != latest_stored_graph:
+            # Store the latest graph into datastream.
```
mitar, line 111: I would add one assert statement here that one hour passed since `latest_stream['latest_datapoint']`.

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: Ah I should point out that I am getting an error if I don't specify start (so just with stream_id and granularity):
`AttributeError: One and only one time range start must be specified.`
But I'm not sure what race condition could occur. Because asking for data does not insert a new stream, right? Can you elaborate?

CdavM: Also, by using `reverse=True` I assume that I would use an earlier `start` parameter and fetch more results and then choose the first one, correct? But why would we fetch more results than necessary (since we're using the default option of `start_exclusive=None`)?

nodewatcher/modules/monitor/http/survey/processors.py
CdavM, line 111: Hmm yes this would prevent many unnecessary reads from the datastream. But then I should check this assertion before I look at the stream, right?

CdavM, line 111: Oh wait you meant to add an assert statement inside that if statement to catch errors in the data collection module? Would this be better than the alternative above?

CdavM, line 111: Hmm OK after thinking about it we'd have to read the previous storage time from the database as well so we're not gaining much. Let's go ahead with your idea.

nodewatcher/modules/monitor/http/survey/processors.py
mitar: OK.

But this is completely wrong. Order of this results is completely arbitrary. It is not necessary that the latest stream for this node is the one you care about. You should use your query tags so that you get precisely at most one stream result back. You can also add assertion that it is at most 1 returned.

You have already set `get_module_name` to configure one of the stream tags you can then use here to select only that stream.

nodewatcher/modules/monitor/http/survey/processors.py
mitar, line 111: Assertion statements are used to catch programming assumptions errors. You are programming under some assumptions. It is good to make those assumptions explicit to catch any errors soon before they can propagate further. For example, an important assumption here is that nodewatcher-agent will give you different survey data not earlier than every one hour (we set granularity to this). If for some reason this is not true it is better that the whole thing goes kaboom instead of failing in an unpredictable way.

mitar, line 111: This has nothing with database reading optimization.

nodewatcher/modules/monitor/http/survey/processors.py
mitar, line 89: OK, it works correctly. See: https://docs.python.org/2/reference/datamodel.html#object.__nonzero__

So it is the same as `len` > 0 because this is how we implemented it: https://github.com/wlanslovenija/datastream/blob/master/datastream/api.py#L237

nodewatcher/modules/monitor/http/survey/processors.py
mitar: > But I'm not sure what race condition could occur. Because asking for data does not insert a new stream, right? Can you elaborate?

You make a query for stream descriptors and then you use `latest_stream['latest_datapoint']` to get the starting point. Now imagine that some other process for some reason (it could be some stray process running data processing in parallel) add one more datapoint to this stream in between. Then `latest_stream['latest_datapoint']` and fetching first entry would not get the latest datapoint, but one before it. Not necessary what you want because it might happen that that datapoint is different to the one you are adding, but the real latest one is not (and you could be inserting too many datapoints in).

mitar: Race condition happens always when you are doing multiple queries to the database. There is always a chance that something will change between your multiple queries. So you have to make queries in a way that this does not matter.

mitar: > Also, by using reverse=True I assume that I would use an earlier start parameter and fetch more results and then choose the first one, correct? But why would we fetch more results than necessary (since we're using the default option of start_exclusive=None)?

No, you use the same `start`, you just use `reverse` so that you know that element at `[0]` is the latest one.

> But why would we fetch more results than necessary (since we're using the default option of start_exclusive=None)?

Also keep in mind that you are not fetching any results until you read them. You just prepare a cursor. You read them when you iterate over them. So `[0]` reads only the first result. (BTW, `len` might be expensive because it might force all elements to be read. I do not remember if we optimized it in this particular case, but in general for iterators `len` might be expensive.)

mitar: In short, you should replicate this query: https://github.com/wlanslovenija/nodewatcher/blob/development/nodewatcher/modules/frontend/topology/static/topology/js/code.js#L11

nodewatcher/modules/monitor/http/survey/processors.py
CdavM, line 111: OK, thanks for the explanation!

I think that the datastream processor is assuming that we always store timestamps according to UTC:
https://github.com/wlanslovenija/nodewatcher/blob/development/nodewatcher/modules/monitor/datastream/processors.py#L73
Should I also assume UTC time to compare the time elapsed?

mitar, line 111: We use timestamps with timezones. If timezone is not available, [UTC timezone is set](https://github.com/wlanslovenija/datastream/blob/master/datastream/api.py#L442).

I think `latest_datapoint` contains a timezone as well. So it does not matter really how you obtain your time, you can just compare timestamps. Python has much better datetime support than JavaScript. So converting to timezones in general is not so important. The only issue is if you do or not use timezones.

You can read more about timezones in Python if you care.

In Django, the best is to use [now](https://github.com/django/django/blob/master/django/utils/timezone.py#L320) function provided by Django. And then just compare with latest datapoint.

CdavM, line 111: > We use timestamps with timezones. If timezone is not available, UTC timezone is set.

So I was trying to say that I think there's a bug which does not use the local timezone for these timestamps, but reverts to UTC. For instance, I set my timezone in https://github.com/wlanslovenija/nodewatcher/blob/development/nodewatcher/settings.py#L48 but my latest datapoint is stored according to UTC (example: `2016-06-21 15:22:17+00:00`)

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: > Now imagine that some other process for some reason (it could be some stray process running data processing in parallel) add one more datapoint to this stream in between. Then latest_stream['latest_datapoint'] and fetching first entry would not get the latest datapoint, but one before it. Not necessary what you want because it might happen that that datapoint is different to the one you are adding, but the real latest one is not (and you could be inserting too many datapoints in).

I am not completely convinced. Let's discuss more. I think you are proposing the following:
1. Survey module stores data at 9am
2. Survey module is running again at 11am, looking for changes. It fetches `latest stream`.
3. Stray module updates the latest datapoint right after the survey module fetches latest stream.
4. Survey module checks the latest datapoint date, which is 11am because of the stray module. It decides that not enough time has passed so it will not update the data.

While I am proposing:
1. Survey module stores data at 9am
2. Survey module is running again at 11am, looking for changes. It fetches `latest stream`.
3. Stray module updates the latest datapoint right after the survey module fetches latest stream.
4. Survey module checks the latest datapoint of the fetched stream, which is 9am, neglecting any actions of the stray module. It decides that enough time has passed so it will update the data.

Or am I confused again?

mitar: > It decides that enough time has passed so it will update the data.

Yes, and so you get to the situation that you updated value twice inside less than an hour.

nodewatcher/modules/monitor/http/survey/processors.py
mitar, line 111: Yes, datapoints will be [always stored as UTC](https://docs.mongodb.com/manual/tutorial/model-time-data/#overview), the question is if the time is correct. So your local time timestamp should be converted to UTC timezone timestamp and then stored. Or are you saying that local timezone is simply dropped and that timestamp is stored with value in your local timezone, but with timezone of UTC, adding a time offset?

mitar, line 111: So stored timestamp should represent the same time as your local time timestamp, only in UTC timezone. If this is not true, then it is a bug. But if you are concerned only about the fact that timezone is not stored, this is not a bug but a feature. We want unified timezone in the database.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,114 @@
+import datetime
+
+from django.utils.translation import gettext_noop as _
+from django.utils import timezone
+from django_datastream import datastream
```
mitar, line 6: Group imports correctly. `django_datastream` is not Django and it is a third party module. Order: core modules, Django modules, nodewatcher modules, and 3rd party modules.

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,114 @@
+import datetime
+
+from django.utils.translation import gettext_noop as _
```
mitar: `django.utils.translation` comes after `django.utils`.

mitar: (In general we are putting `_` always last.)

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def get_module_name(self):
+        return 'monitor.http.survey'
+
+    def get_stream_highest_granularity(self):
+        # The nodewatcher-agent performs a survey once every ~240 monitoring intervals according to
+        # https://github.com/wlanslovenija/nodewatcher-agent/blob/master/modules/wireless.c#L362
+        # and a monitoring run is performed every 30 seconds according to
+        # https://github.com/wlanslovenija/nodewatcher-agent/blob/73f2b25db2c34ae7d70904bac31379a2243bade0/modules/wireless.c#L437
+        # (master branch at https://github.com/wlanslovenija/nodewatcher-agent/blob/master/modules/wireless.c#L437).
```
mitar: Just master branch, no need for both. Why both?

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,114 @@
...
+        latest_stored_graph = None
+        # Retrieve the latest stored datapoint, if it exists
+        streams = datastream.find_streams({'node': node.uuid, 'module': 'monitor.http.survey'})
+        if streams:
+            assert len(streams) == 1
+            datapoint_iterator = datastream.get_data(
+                stream_id=streams[0]['stream_id'],
+                granularity=streams[0]['highest_granularity'],
+                start=streams[0]['latest_datapoint'],
+                reverse=True)
```
mitar: `reverse=True)` ->

```
    reverse=True,
)
```

nodewatcher/modules/monitor/http/survey/processors.py
CdavM, line 111: > So stored timestamp should represent the same time as your local time timestamp, only in UTC timezone.

This does happen, so I think it's OK. I was wondering why you'd want to store values into the database according to the UTC timezone instead of the one specified in the settings.

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: But why do you think that the same value gets updated by the survey module and the stray module? Or are you saying that two workers can run the same module at the same time?

nodewatcher/modules/monitor/http/survey/processors.py
mitar, line 111: Setting is what should be used to display timestamps by default. You want to convert timestamps when displaying to the user anyway.

(In our case, we convert timestamps to the timezone of a node, and not user, when displaying data. And you can imagine that when nodes are global, all across the world, one setting in the database is not enough.)

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -0,0 +1,114 @@
...
+        latest_graph = {
+            'v': vertices,
+            'e': edges
+        }
+
+        if latest_graph != latest_stored_graph:
+            # Since a new survey is performed once every two hours, it should be impossible to insert new data more
+            # often than once every hour.
+            if latest_stored_graph:
+                assert timezone.now() - streams[0]['latest_datapoint'] > datetime.timedelta(hours=1)
```
mitar:
```
assert not latest_stored_graph or timezone.now() - streams[0]['latest_datapoint'] >= datetime.timedelta(hours=1)
```

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: I was thinking that the master branch can change, while that specific commit is there to stay. What if something else will be at line 437 in a week?

mitar: A agree. In general I would prefer something like:

> The nodewatcher-agent performs a survey once every ~240 monitoring intervals and a monitoring run is performed every 30 seconds according to the wireless module of nodewatcher-agent.

Because even links to the code change. And implementations as well.

What do you think?

nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -108,8 +108,9 @@ def process(self, context, node):
         if latest_graph != latest_stored_graph:
             # Since a new survey is performed once every two hours, it should be impossible to insert new data more
             # often than once every hour.
-            if latest_stored_graph:
-                assert timezone.now() - streams[0]['latest_datapoint'] > datetime.timedelta(hours=1)
+            assert not latest_stored_graph or timezone.now() - streams[0]['latest_datapoint'] >= datetime.timedelta(
+                hours=1,
+            )
```
mitar: 120 is not a strict limit. In this case for example I would not do it myself. :-)

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: Okay, I think that the wireless module of nodewatcher-agent is easy enough to find. But should we at least say that the frequency of a run is currently defined as the refresh interval?

nodewatcher/modules/monitor/http/survey/processors.py
CdavM: Changed back :-)

nodewatcher/modules/monitor/http/survey/processors.py
mitar: Sure.

mitar:
I think this is good now. @kostko, what do you think? Ready to merge?

mitar:
BTW, you are writing documentation on how to write nodewatcher modules in meantime? :-)

kostko:
I think these commits should be squashed into a single commit called "Implement survey module." or something. And rebased on latest development branch. Because most of these are just fixes based on review comments.

mitar:
Hm, we have never done squashing and rebasing? I am against loosing history. Why we would start now? It does not matter if it is ugly, this was also a process.

But otherwise you are OK with the content of the pull request?

mitar:
OK, I am merging it as it is. :-)

mitar:
Congrats!

PR #23, opened 2016-07-06
Export survey module data

CdavM:
This is a command that exports a graph of the survey. It currently only exports data from the latest datapoint. This PR should be used to discuss implementation details and additional features for this script.

### DIFF
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -34,6 +34,7 @@ class Command(base.BaseCommand):
         """
         Exports the latest survey data graph as a JSON file into the root directory.
         """
+
         number = options['number']
         label = options['label']
 
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -81,7 +81,7 @@ def extract_survey_graph(upper_datetime):
 
     streams = datastream.find_streams({'module': 'monitor.http.survey'})
 
-    # Create a meta graph
+    # Create a meta graph consolidating date from all streams.
     meta_vertices, meta_edges = [], []
     friendly_nodes = []
     latest_data_timepoint = None
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -19,17 +19,17 @@ class Command(base.BaseCommand):
             type=int,
             action='store',
             dest='timestamp',
-            default=int(time.time()),
             help="All the data will be collected between this UNIX timestamp and two hours preceding it."
         )
 
         parser.add_argument(
-            '--label',
+            '-f',
+            '--filename',
             type=str,
             action='store',
-            dest='label',
+            dest='filename',
             default='survey_export',
-            help="Optional label for the filename",
+            help="Optional filename",
         )
 
         parser.add_argument(
@@ -48,7 +48,7 @@ class Command(base.BaseCommand):
 
         timestamp = options['timestamp']
         date_from_timestamp = datetime.datetime.fromtimestamp(timestamp)
-        label = options['label']
+        filename = options['filename']
 
         streams = datastream.find_streams({'module': 'monitor.http.survey'})
 
@@ -88,9 +88,9 @@ class Command(base.BaseCommand):
         }
 
         if options['store_timestamp']:
-            filename = '{0}{1}.json'.format(label, timestamp)
+            filename = '{0}{1}.json'.format(filename, timestamp)
         else:
-            filename = '{0}.json'.format(label)
+            filename = '{0}.json'.format(filename)
 
         with io.open(filename, 'w', encoding='utf-8') as f:
             f.write(json.dumps(exported_graph, ensure_ascii=False))
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -59,6 +59,8 @@ class Command(base.BaseCommand):
         # Create a meta graph
         meta_vertices, meta_edges = [], []
         friendly_nodes = []
+        latest_data_timepoint = 0
+
         for stream in streams:
             datapoint_iterator = datastream.get_data(
                 stream_id=stream['stream_id'],
@@ -74,6 +76,8 @@ class Command(base.BaseCommand):
                         friendly_nodes.append(vertex)
                 for edge in stream_graph['e']:
                     meta_edges.append(edge)
+                if not latest_data_timepoint or datapoint_iterator[0]['t'] > latest_data_timepoint:
+                    latest_data_timepoint = datapoint_iterator[0]['t']
             except IndexError:
                 pass
 
@@ -89,6 +93,7 @@ class Command(base.BaseCommand):
         exported_graph = {
             'graph': meta_graph,
             'friendly_nodes': friendly_nodes,
+            'timepoint': str(latest_data_timepoint)
         }
 
         if options['store_timestamp']:
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -52,14 +52,13 @@ class Command(base.BaseCommand):
         try:
             upper_datetime = datetime.datetime.strptime(string_upper_datetime, '%Y-%m-%dT%H:%M:%S')
         except ValueError:
-            self.stdout.write(self.style.ERROR("Unable to parse the date parameter."))
-            return
+            raise base.CommandError("Unable to parse the date parameter.")
 
         exported_graph = extract_survey_graph(upper_datetime)
 
         if not basename:
             # Redirect to stdout.
-            self.stdout.write(exported_graph,  ending="")
+            self.stdout.write(exported_graph, ending="")
             return
 
         filename = '{0}.json'.format(basename)
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -56,8 +56,7 @@ class Command(base.BaseCommand):
         exported_graph = extract_survey_graph(parsed_datetime)
 
         if not exported_graph:
-            self.stdout.write(self.style.ERROR("Insufficient survey data in the datastream during this time period."))
-            return
+            raise base.CommandError("Insufficient survey data in the datastream during this time period.")
 
         filename = '{0}.json'.format(filename)
 
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -59,9 +59,6 @@ class Command(base.BaseCommand):
 
         exported_graph = extract_survey_graph(parsed_datetime)
 
-        if not exported_graph:
-            raise base.CommandError("Insufficient survey data in the datastream during this time period.")
-
         filename = '{0}.json'.format(filename)
 
         with io.open(filename, 'w', encoding='utf-8') as f:
@@ -115,7 +112,7 @@ def extract_survey_graph(parsed_datetime=None):
             pass
 
     if not meta_vertices or not meta_edges:
-        return
+        raise LookupError("Insufficient survey data in the datastream during this time period.")
 
     meta_graph = {
         'v': meta_vertices,
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -37,7 +37,7 @@ class Command(base.BaseCommand):
             type=bool,
             action='store',
             dest='store_timestamp',
-            default=True,
+            default=False,
             help="Include timestamp in filename?",
         )
 
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -8,17 +8,20 @@ from django_datastream import datastream
 
 
 class Command(base.BaseCommand):
-    help = "Consolidates survey data from all nodes and exports the meta graph as a JSON file."
+    help = "Consolidates the latest survey data within the specified time interval \
+           from all nodes and exports the meta graph as a JSON file."
 
     def add_arguments(self, parser):
 
         # Optional arguments
         parser.add_argument(
-            '--timestamp',
-            type=int,
+            '--datetime',
+            type=str,
             action='store',
-            dest='timestamp',
-            help="All the data will be collected between this UNIX timestamp and two hours preceding it."
+            dest='datetime',
+            help="All the data will be collected between this UTC datetime \
+                 (format 'yyyy-mm-ddThh:mm:ss' (i.e. '2007-03-04T12:00:00 2007-04-10T12:00:00'), \
+                  and two hours preceding it."
         )
 
         parser.add_argument(
@@ -32,12 +35,12 @@ class Command(base.BaseCommand):
         )
 
         parser.add_argument(
-            '--include-timestamp-in-filename',
+            '--include-datetime-in-filename',
             type=bool,
             action='store',
-            dest='store_timestamp',
+            dest='store_datetime',
             default=False,
-            help="Include timestamp in filename?",
+            help="Include datetime in filename?",
         )
 
     def handle(self, *args, **options):
@@ -45,17 +48,25 @@ class Command(base.BaseCommand):
         Exports the latest survey data graph as a JSON file into the root directory.
         """
 
-        timestamp = options['timestamp']
+        parsed_datetime = None
+        input_datetime = options['datetime']
         filename = options['filename']
 
-        exported_graph = extract_survey_graph(timestamp)
+        if input_datetime:
+            try:
+                parsed_datetime = datetime.datetime.strptime(options['datetime'], "%Y-%m-%dT%H:%M:%S")
+            except ValueError:
+                self.stdout.write(self.style.ERROR("Unable to parse datetime."))
+                return
+
+        exported_graph = extract_survey_graph(parsed_datetime)
 
         if not exported_graph:
             self.stdout.write(self.style.ERROR("Insufficient survey data in the datastream during this time period."))
             return
 
-        if options['store_timestamp']:
-            filename = '{0}{1}.json'.format(filename, timestamp)
+        if options['store_datetime']:
+            filename = '{0}{1}.json'.format(filename, str(datetime))
         else:
             filename = '{0}.json'.format(filename)
 
@@ -65,17 +76,17 @@ class Command(base.BaseCommand):
         self.stdout.write(self.style.SUCCESS("Successfully exported survey data."))
 
 
-def extract_survey_graph(timestamp=None):
+def extract_survey_graph(parsed_datetime=None):
     """
     Extracts a graph of the survey during a specified time period. Only the latest data
     from specified datetime and two hours preceding the datetime will be used.
 
-    :param timestamp: UNIX timestamp according to the UTC timezone.
+    :param parsed_datetime: Datetime object according to the UTC timezone.
     :return: A dictionary that contains the graph under the 'graph' key.
     """
 
-    if timestamp:
-        latest_survey_datetime = datetime.datetime.fromtimestamp(timestamp)
+    if parsed_datetime:
+        latest_survey_datetime = parsed_datetime
     else:
         latest_survey_datetime = datetime.datetime.utcnow()
 
@@ -91,6 +102,7 @@ def extract_survey_graph(timestamp=None):
             stream_id=stream['stream_id'],
             granularity=stream['highest_granularity'],
             start=(latest_survey_datetime - datetime.timedelta(hours=2)),
+            end=latest_survey_datetime,
             reverse=True,
         )
         try:
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -46,54 +46,13 @@ class Command(base.BaseCommand):
         """
 
         timestamp = options['timestamp']
-        if timestamp:
-            latest_survey_datetime = datetime.datetime.fromtimestamp(timestamp)
-        else:
-            latest_survey_datetime = datetime.datetime.utcnow()
-
         filename = options['filename']
 
-        streams = datastream.find_streams({'module': 'monitor.http.survey'})
+        exported_graph = extract_survey_graph(timestamp)
 
-        # Create a meta graph
-        meta_vertices, meta_edges = [], []
-        friendly_nodes = []
-        latest_data_timepoint = 0
-
-        for stream in streams:
-            datapoint_iterator = datastream.get_data(
-                stream_id=stream['stream_id'],
-                granularity=stream['highest_granularity'],
-                start=(latest_survey_datetime - datetime.timedelta(hours=2)),
-                reverse=True,
-            )
-            try:
-                stream_graph = datapoint_iterator[0]['v']
-                for vertex in stream_graph['v']:
-                    meta_vertices.append(vertex)
-                    if 'b' in vertex:
-                        friendly_nodes.append(vertex)
-                for edge in stream_graph['e']:
-                    meta_edges.append(edge)
-                if not latest_data_timepoint or datapoint_iterator[0]['t'] > latest_data_timepoint:
-                    latest_data_timepoint = datapoint_iterator[0]['t']
-            except IndexError:
-                pass
-
-        if not meta_vertices or not meta_edges:
+        if not exported_graph:
             self.stdout.write(self.style.ERROR("Insufficient survey data in the datastream during this time period."))
             return
-
-        meta_graph = {
-            'v': meta_vertices,
-            'e': meta_edges,
-        }
-
-        exported_graph = {
-            'graph': meta_graph,
-            'friendly_nodes': friendly_nodes,
-            'timepoint': str(latest_data_timepoint)
-        }
 
         if options['store_timestamp']:
             filename = '{0}{1}.json'.format(filename, timestamp)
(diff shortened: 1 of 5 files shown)


nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        timestamp = int(time.time())
+
+        if label:
+            filename = label + str(timestamp) + '.json'
+        else:
+            filename = 'survey_export' + str(timestamp) + '.json'
+
+        with io.open(filename, 'w', encoding='utf-8') as f:
```
CdavM, line 66: What would be a good place to store these files?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class Command(base.BaseCommand):
+    help = 'Consolidates survey data from all nodes and exports the meta graph as a JSON file.'
+
+    def add_arguments(self, parser):
+
+        # Optional arguments
+        parser.add_argument(
+            '--surveys-to-export',
```
CdavM: Better name for this parameter?

CdavM:
Right now we're simply adding all vertices and edges into one big graph. We could also create a networkX graph structure and export that. Would this be better?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,81 @@
...
+        meta_graph = {
+            'v': meta_vertices,
+            'e': meta_edges,
+        }
+        exported_graph = {
+            'graph': meta_graph,
+            'friendly_nodes': friendly_nodes,
+        }
+
+        timestamp = int(time.time())
```
CdavM: The timestamp should be the time data was created, not when it was exported.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
mitar: What should this be? You export only one common json for all nodes/surveys, but only the timestamp (which time) you configure, no?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,81 @@
...
+import json
+import time
+
+from django.core.management import base
+
+from django_datastream import datastream
+
+
+class Command(base.BaseCommand):
+    help = 'Consolidates survey data from all nodes and exports the meta graph as a JSON file.'
```
mitar: User facing strings should use `"`.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,81 @@
...
+    def add_arguments(self, parser):
+
+        # Optional arguments
+        parser.add_argument(
+            '--surveys-to-export',
+            type=int,
+            action='store',
+            dest='number',
+            default=1,
+            help='Number of surveys to export'
```
mitar: User facing strings should use `"`. Comma at the end.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        Exports the latest survey data graph as a JSON file into the root directory.
+        """
+
+        number = options['number']
+        label = options['label']
+
+        streams = datastream.find_streams({'module': 'monitor.http.survey'})
+
+        # Create a meta graph
```
mitar: So this code should probably not be duplicated later on in your module.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
mitar: Yes, so? Use `stream['latest_datapoint']`. Or average of those.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
mitar, line 66: I think current directory is good.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,81 @@
...
+            'e': meta_edges,
+        }
+        exported_graph = {
+            'graph': meta_graph,
+            'friendly_nodes': friendly_nodes,
+        }
+
+        timestamp = int(time.time())
+
+        if label:
```
mitar: This should probably be simply a default for label argument.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,81 @@
...
+            filename = label + str(timestamp) + '.json'
+        else:
+            filename = 'survey_export' + str(timestamp) + '.json'
```
mitar: Use string interpolation for making strings, not concatenation.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,81 @@
...
+
+        with io.open(filename, 'w', encoding='utf-8') as f:
+            f.write(json.dumps(exported_graph, ensure_ascii=False))
+            self.stdout.write(self.style.SUCCESS("Successfully exported survey data."))
```
mitar: This can be outside of `with`.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
mitar, line 66: It is pretty common to allow user to simply specify filename/path as an argument and this is it. You want also to specify the timestamp into the filename and this makes it more complicated. Is timestamp really necessary?

I would maybe make it so that you have one argument which is the timestamp for which you want to get data, and if it is not specified, current time is used. So, user/caller knows the timestamp (or it is given, or it is current time) and can use Bash or something else to construct the filename if they care about having the timestamp in the filename.

So maybe simplify this so that simply it is just a filename.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,81 @@
...
+        number = options['number']
+        label = options['label']
+
+        streams = datastream.find_streams({'module': 'monitor.http.survey'})
+
+        # Create a meta graph
+        meta_vertices, meta_edges = [], []
+        friendly_nodes = []
+        for stream in streams:
+            datapoint_iterator = datastream.get_data(
```
mitar: I am not sure if this is the right approach. So you should find surveys around a given timestamp, even if this is the current timestamp.

Imagine that you would have a node which would go down for few days. You do not want to include its survey data from two days ago into this network here. So you should just find [timestamp, timstamp - 2 hours] surveys for a given timestamp and use only those.

CdavM: Hmm yes this is a good point. I had two uses in mind for this tool:
1. To be able to export the latest data for debugging
2. To archive graphs. This is motived by my current data collection project. So I'd like to move the rogue node 10 times and then say: find the 10 latest graphs and export them into JSON files.

How should we proceed? Is there a good way of implementing the second option?

mitar: You could have a bash for loop which goes and creates timestamps for every two hours and gives them to this script? Or you would really like to search for latest n different surveys.

If you need that, then feel free to do it. Then make timestamp + how many different survey graphs you want from that timestamp into the past.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,125 @@
...
+            reverse=True,
+        )
+        try:
+            stream_graph = datapoint_iterator[0]['v']
+            for vertex in stream_graph['v']:
+                meta_vertices.append(vertex)
+                if 'b' in vertex:
+                    friendly_nodes.append(vertex['i'])
+                    for bssid in vertex['b']:
+                        friendly_nodes.append(bssid)
```
CdavM, line 103: How should we construct this friendly_nodes list? This structure is most convenient for my algorithm but maybe not very legible.

mitar, line 103: Whatever it is good for you for now.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,125 @@
...
+        else:
+            filename = '{0}.json'.format(filename)
+
+        with io.open(filename, 'w', encoding='utf-8') as f:
+            f.write(json.dumps(exported_graph, ensure_ascii=False))
+
+        self.stdout.write(self.style.SUCCESS("Successfully exported survey data."))
+
+
+def extract_survey_graph(timestamp=None):
```
mitar: Use `datetime` objects as arguments.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def add_arguments(self, parser):
+
+        # Optional arguments
+        parser.add_argument(
+            '--timestamp',
+            type=int,
+            action='store',
+            dest='timestamp',
+            help="All the data will be collected between this UNIX timestamp and two hours preceding it."
```
mitar: But only one datapoint in this interval, no? So the closest one preceding it?

CdavM: Yes, only the latest one. But the timestamp ensures any data collected falls within the specified interval.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
CdavM: Fixed with 4c7c52e.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
mitar: Yes, I know. I am just saying that the text is slightly unclear about that.

CdavM:
OK, is this ready to be merged?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+import json
+import datetime
+
+from django.core.management import base
+
+from django_datastream import datastream
+
+
+class Command(base.BaseCommand):
+    help = "Consolidates the latest survey data within the specified time interval \
```
mitar: Use `"""` for such long multi-line strings.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+           from all nodes and exports the meta graph as a JSON file."
+
+    def add_arguments(self, parser):
+
```
mitar: No empty line.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+        # Optional arguments
+        parser.add_argument(
+            '--datetime',
+            type=str,
+            action='store',
+            dest='datetime',
+            help="All the data will be collected between this UTC datetime \
+                 (format 'yyyy-mm-ddThh:mm:ss' (i.e. '2007-03-04T12:00:00 2007-04-10T12:00:00'), \
+                  and two hours preceding it."
```
mitar: Indentation is not equal between lines.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+            '--filename',
+            type=str,
+            action='store',
+            dest='filename',
+            default='survey_export',
+            help="Optional filename",
+        )
+
+        parser.add_argument(
+            '--include-datetime-in-filename',
```
mitar: I would change `datetime` to `timestamp` here. So you are storing timestamp, but internally it is represented as a datetime. So just names in this strings is what I am commenting here.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+            type=bool,
+            action='store',
+            dest='store_datetime',
+            default=False,
+            help="Include datetime in filename?",
```
mitar: Maybe explain in which way it is included in the filename? Appended? Example?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+        parsed_datetime = None
+        input_datetime = options['datetime']
+        filename = options['filename']
+
+        if input_datetime:
+            try:
+                parsed_datetime = datetime.datetime.strptime(options['datetime'], "%Y-%m-%dT%H:%M:%S")
+            except ValueError:
+                self.stdout.write(self.style.ERROR("Unable to parse datetime."))
+                return
```
mitar: Hm, can you return error in some way? So that return value of the manage script is non-zero to signal an error in Bash?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+    def add_arguments(self, parser):
+
+        # Optional arguments
+        parser.add_argument(
+            '--datetime',
+            type=str,
+            action='store',
+            dest='datetime',
+            help="All the data will be collected between this UTC datetime \
+                 (format 'yyyy-mm-ddThh:mm:ss' (i.e. '2007-03-04T12:00:00 2007-04-10T12:00:00'), \
```
mitar: I think you copy-pasted too much here. :-) Example contains two timestamps.

Should timestamp really be provided in UTC? Do you really parse it as UTC?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+
+class Command(base.BaseCommand):
+    help = "Consolidates the latest survey data within the specified time interval \
+           from all nodes and exports the meta graph as a JSON file."
+
+    def add_arguments(self, parser):
+
+        # Optional arguments
+        parser.add_argument(
+            '--datetime',
```
mitar: `timestamp` is better, `datetime` is internal implementation detail. End-users do not care how classes are named in Python.

You could also rename this argument to `at` or something like that.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+            type=str,
+            action='store',
+            dest='datetime',
+            help="All the data will be collected between this UTC datetime \
```
mitar: Comment is still unclear what this argument really do.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+
+        parser.add_argument(
+            '-f',
+            '--filename',
+            type=str,
+            action='store',
+            dest='filename',
+            default='survey_export',
+            help="Optional filename",
```
mitar: Dot is missing at the end. In which way it is optional. Maybe specify example. What if it is not provided, how will filename look like? Or will the output go to stdout?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+            help="Include datetime in filename?",
+        )
+
+    def handle(self, *args, **options):
+        """
+        Exports the latest survey data graph as a JSON file into the root directory.
+        """
+
+        parsed_datetime = None
+        input_datetime = options['datetime']
```
mitar: Keep variable names the same as key names. Do not rename things like that because it is harder to read the code. Name this option better from the beginning.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+        filename = options['filename']
+
+        if input_datetime:
+            try:
+                parsed_datetime = datetime.datetime.strptime(options['datetime'], "%Y-%m-%dT%H:%M:%S")
```
mitar: Are you really parsing this in UTC?

You should probably use this Django functions anyway: https://github.com/django/django/blob/master/django/utils/dateparse.py

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+    from specified datetime and two hours preceding the datetime will be used.
+
+    :param parsed_datetime: Datetime object according to the UTC timezone.
+    :return: A dictionary that contains the graph under the 'graph' key.
+    """
+
+    if parsed_datetime:
+        latest_survey_datetime = parsed_datetime
+    else:
+        latest_survey_datetime = datetime.datetime.utcnow()
```
mitar: You should probably use: https://github.com/django/django/blob/master/django/utils/timezone.py#L320

Also, this argument checking should probably be part of the caller code. API here should simply require that timestamp argument is provided. Also, why `parsed_datetime` as argument name? From the perspective of this function it is completely unnecessary to know how was timestamp obtained.

On the other hand it is also bad to change the argument name between caller and callee. Variable passed to the function should be called the same as an argument, if possible. So the best is to think how the argument here should be called to be clear what it is, and then use it everywhere, including arguments.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+            try:
+                parsed_datetime = datetime.datetime.strptime(options['datetime'], "%Y-%m-%dT%H:%M:%S")
+            except ValueError:
+                self.stdout.write(self.style.ERROR("Unable to parse datetime."))
+                return
+
+        exported_graph = extract_survey_graph(parsed_datetime)
+
+        if not exported_graph:
+            self.stdout.write(self.style.ERROR("Insufficient survey data in the datastream during this time period."))
```
mitar: Wouldn't it be better if `extract_survey_graph` would throw an exception with such message? Because how does really caller know what is the reason why `extract_survey_graph` could not finish?

Use exceptions for errors. This is why they are. I think there is even a special class for manage script exceptions you can use.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
mitar: Also, constant string should use `'`.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+            return
+
+        if options['store_datetime']:
+            filename = '{0}{1}.json'.format(filename, str(datetime))
+        else:
+            filename = '{0}.json'.format(filename)
```
mitar: Filename is understood to be the whole filename. You should consider simply using whatever user provided.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
mitar: Is default printed out automatically?

Also, sometimes user might want to output to stdout. One way to allow this is that if filename is not specified, it goes to stdout. Another is to specially handle filename `-`. If it is `-` you send it to stdout. You should decide which one you prefer and document it.

It is pretty common that if file is not provided it goes to stdout. Not that a filename is just picked. (Because then you might override something by accident.)

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+                return
+
+        exported_graph = extract_survey_graph(parsed_datetime)
+
+        if not exported_graph:
+            self.stdout.write(self.style.ERROR("Insufficient survey data in the datastream during this time period."))
+            return
+
+        if options['store_datetime']:
+            filename = '{0}{1}.json'.format(filename, str(datetime))
```
mitar: I think it would be better if you would parse the provided filename, extract extension, and then append timestamp, so that the result is something like `<basename><timestamp><extension>`.

mitar: Python provides utilities for things like that.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+        latest_survey_datetime = parsed_datetime
+    else:
+        latest_survey_datetime = datetime.datetime.utcnow()
+
+    streams = datastream.find_streams({'module': 'monitor.http.survey'})
+
+    # Create a meta graph
+    meta_vertices, meta_edges = [], []
+    friendly_nodes = []
+    latest_data_timepoint = 0
```
mitar: `0` or `None`?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+                meta_edges.append(edge)
+            if not latest_data_timepoint or datapoint_iterator[0]['t'] > latest_data_timepoint:
+                latest_data_timepoint = datapoint_iterator[0]['t']
+        except IndexError:
+            pass
+
+    if not meta_vertices or not meta_edges:
+        return
+
+    meta_graph = {
```
mitar, line 114: What does `meta` mean in the name of this variable?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -0,0 +1,137 @@
...
+        'v': meta_vertices,
+        'e': meta_edges,
+    }
+
+    exported_graph = {
+        'graph': meta_graph,
+        'friendly_nodes': friendly_nodes,
+        'timepoint': str(latest_data_timepoint)
```
mitar, line 122: `timestamp`?

mitar, line 122: Do you have to convert it to string? JSON can serialize datetime, or not?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
CdavM: It should be in UTC because all timestamps are stored in UTC.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
CdavM: I actually used `now()` before. But I realized that timestamps are always stored according to UTC and now() does not return a time in UTC. Which makes this pointless. We need to either store timestamps according to the local timezone or use UTC times to browse through timepoints.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
CdavM, line 122: `TypeError: datetime.datetime(2016, 7, 18, 5, 18, 51, tzinfo=<bson.tz_util.FixedOffset object at 0x7f0519ff0f50>) is not JSON serializable`
:-(

CdavM:
I made almost all of the changes requested. I am still going to ask users to enter the date in a human-readable format. Otherwise they have to enter the date somewhere to get the timestamp out, which is really inconvenient. I hid the implementation details about the datetime module from users. I am still using UTC. I am using the datetime module from python instead of django_utils because datetime is more common + used elsewhere in nodewatcher.

CdavM:
OK I now added an option to feed the data to the rogue node detection algorithm. But this requires an import statement to the rogue node algorithm. Which I can't commit because the rogue node algorithm hasn't been merged yet. So I think we should first merge this, then work on merging the rogue node detection and then I can push a final commit here.

CdavM:
Can I get another code review?

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
mitar: It does not matter in which way they are stored. The whole point of using timezone-aware timestamps is that Python knows how to convert between them.

But this should not matter at all if you are using Django date parsing anyway.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
mitar: I think you should read more about timestamps in Python. You can have timezone aware and non-aware ones. It does not matter how timestamps are stored. Timezones are useful for users, so that they do not have to think in other timezone but can provide their local timezone. We are storing data in a normalized timezone in a database so that it is simpler to store. But when dealing with users it is nice to adapt to what they use.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
mitar, line 122: This is probably why you should use Django JSON serializer: https://github.com/django/django/blob/master/django/core/serializers/json.py

mitar:
> I am still going to ask users to enter the date in a human-readable format.

Yes, that was the whole point. You ask them in human-readable format (where allowing them to enter timestamp in local timezone is additional thing which makes them even more user-friendly, with optional timezone), but then you quickly convert this to datetime and you use that all around Python.

> I am using the datetime module from python instead of django_utils because datetime is more common + used elsewhere in nodewatcher.

???

Those things are complementary. Nodewatcher is written in Django, so using Django-provided functions help make it better. For example, it allows one to use settings to configure some aspects of data parsing and formatting and you can then through settings decide if you want or do not want to use timezones.

mitar:
I made a new pull request from a branch in this repository so that I can simply fix some things instead of going through long comments iterations.

I think you made great stuff, but I just changed some wording to make it clearer (in my personal opinion).

Please check that pull request and if you are OK with it I will merge it.

mitar:
I have also made you a contributor to this repository. So from now on create branches inside this repository. (But you cannot push to `development` branch directly. And you should not. Still create pull requests to the `development` branch.)

PR #24, opened 2016-07-08
Removed the assertion in the survey module.

CdavM:


nodewatcher/modules/monitor/http/survey/processors.py
```diff
@@ -106,13 +103,12 @@ def process(self, context, node):
...
-            'e': edges
+            'e': edges,
         }
 
         if latest_graph != latest_stored_graph:
-            # Since a new survey is performed once every two hours, it should be impossible to insert new data more
-            # often than once every hour.
-            assert not latest_stored_graph or timezone.now() - streams[0]['latest_datapoint'] >= datetime.timedelta(hours=1)
+            # Since a new survey is performed once every two hours,
+            # new data should only be inserted once in that time period.
```
mitar, line 112: Maybe also explain that it is not a problem if it is done more often, because the maximum granularity setting is only an optimization.

mitar, line 112: And no data is lost.

mitar:
LGTM.

But do you know why surveys are more often than one hour? We should debug that as well.

PR #26, opened 2016-07-22
Export survey module data

mitar:
A followup from #23 with branch made in this repository.

### DIFF
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -34,6 +34,7 @@ class Command(base.BaseCommand):
         """
         Exports the latest survey data graph as a JSON file into the root directory.
         """
+
         number = options['number']
         label = options['label']
 
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -81,7 +81,7 @@ def extract_survey_graph(upper_datetime):
 
     streams = datastream.find_streams({'module': 'monitor.http.survey'})
 
-    # Create a meta graph
+    # Create a meta graph consolidating date from all streams.
     meta_vertices, meta_edges = [], []
     friendly_nodes = []
     latest_data_timepoint = None
--- a/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
+++ b/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
@@ -19,17 +19,17 @@ class Command(base.BaseCommand):
             type=int,
             action='store',
             dest='timestamp',
-            default=int(time.time()),
             help="All the data will be collected between this UNIX timestamp and two hours preceding it."
         )
 
         parser.add_argument(
-            '--label',
+            '-f',
+            '--filename',
             type=str,
             action='store',
-            dest='label',
+            dest='filename',
             default='survey_export',
-            help="Optional label for the filename",
+            help="Optional filename",
         )
 
         parser.add_argument(
@@ -48,7 +48,7 @@ class Command(base.BaseCommand):
 
         timestamp = options['timestamp']
         date_from_timestamp = datetime.datetime.fromtimestamp(timestamp)
-        label = options['label']
+        filename = options['filename']
 
         streams = datastream.find_streams({'module': 'monitor.http.survey'})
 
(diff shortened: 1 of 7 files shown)


mitar:
@CdavM: Please test if it works. I have not run the code with my changes. If you notice anything, feel free to just fix it.

CdavM:
@mitar : The code doesn't work anymore :( 
When I try running your command to export the data, I get the following error:

```
Traceback (most recent call last):
  File "manage.py", line 9, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/code/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py", line 68, in handle
    json.dump(graph, f, ensure_ascii=False, cls=django_json.DjangoJSONEncoder)
  File "/usr/lib/python2.7/json/__init__.py", line 190, in dump
    fp.write(chunk)
TypeError: must be unicode, not str
```

After reverting to f.write(json.dumps(...)), I get the following error:

```
Traceback (most recent call last):
  File "manage.py", line 9, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/code/nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py", line 68, in handle
    f.write(json.dumps(graph, ensure_ascii=False))
  File "/usr/lib/python2.7/json/__init__.py", line 250, in dumps
    sort_keys=sort_keys, **kw).encode(obj)
  File "/usr/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: datetime.datetime(2016, 7, 25, 4, 7, 48, tzinfo=<bson.tz_util.FixedOffset object at 0x7fb1a377bf90>) is not JSON serializable
```

So I'd just go back to my way of exporting. I have no idea how to track down these unicode errors. Or do you have any ideas on fixing this?

CdavM:
OK I fixed writing the output and I reverted to using 'v' and 'e' as keys, as that is done elsewhere in the code.

nodewatcher/modules/monitor/http/survey/management/commands/export_survey_data.py
```diff
@@ -136,8 +136,8 @@ def all_nodes_survey_graph(at):
 
     exported_graph = {
         'graph': {
-            'vertices': vertices,
-            'edges': edges,
+            'v': vertices,
+            'e': edges,
```
mitar, line 133: Why this? Those files are not size-sensitive. We are compressing keys in the database because that is an optimization. Here we do not need to optimize anything.

CdavM, line 133: Yes but we're otherwise breaking an abstraction barrier.

The rogue detection algorithm should not need to know if it's processing data that comes from just one stream (in which case the keys are 'v' and 'e') or if the data came from this function that combines data from multiple streams (in which case the keys are 'vertices' and 'edges').

So it should always rely on the same key, hence we should revert to 'v' and 'e'. This doesn't matter for the other two fields (known_nodes and timestamp)..

mitar, line 133: Good thinking! Maybe then add a comment here explaining this.

mitar:
Above you mentioned that you had issues exporting dates, but then I do not see that you reverted to your old way of exporting dates as strings. So it works for you as it is now?

PR #27, opened 2016-07-25
Rogue node detection.

CdavM:
This PR should be used to discuss an implementation of the rogue node algorithm implementation.


mitar:
So, why are tests failing here?

nodewatcher/modules/analysis/rogue_nodes/tasks.py
```diff
@@ -0,0 +1,81 @@
+import datetime
+
+from django.core.mail import mail_admins
+
+from django.conf import settings
+
+from nodewatcher import celery
+
+from nodewatcher.modules.monitor.http.survey.management.commands.export_survey_data import extract_survey_graph
```
mitar: I think it would be better if this particular function would be somewhere else, and then both the management command and this task would import it.

nodewatcher/modules/analysis/rogue_nodes/tasks.py
```diff
@@ -0,0 +1,81 @@
+import datetime
+
+from django.core.mail import mail_admins
```
mitar: Import only modules, never methods/functions directly.

nodewatcher/modules/analysis/rogue_nodes/tasks.py
```diff
@@ -0,0 +1,81 @@
...
+
```
mitar: No newlines between imports of the same top-level package.

CdavM:
It's no longer failing. Had to rename a couple of modules.

nodewatcher/modules/analysis/rogue_nodes/tests.py
```diff
@@ -0,0 +1,28 @@
...
+                # Test every json file that does not contain '-results' in the filename
+                if filename.split('.')[-1] == 'json' and 'results' not in filename:
+                    # Run the algorithm on the file and compare the results.
+                    print('testing ' + str(filename))
+                    input_graph = json.load(io.open(os.path.join(path, filename), encoding='utf-8'))
+                    output_results = rogue_node_detection_algorithm(
+                        graph=input_graph['graph'],
+                        friendly_nodes=input_graph['friendly_nodes'],
+                    )
+                    # append -results to the end of the filename
```
mitar: Comment style.

nodewatcher/modules/analysis/rogue_nodes/tests.py
```diff
@@ -0,0 +1,28 @@
+from django import test as django_test
+
+from nodewatcher.modules.analysis.rogue_nodes.tasks import rogue_node_detection_algorithm
+
+import json
+import io
```
mitar, line 1: Imports order.

mitar:
Hm, do you think it is useful to store whole results for your tests? Maybe. But isn't the idea that you detect rogue nodes, but slight changes in values might not be so problematic? So maybe more important is that class rogue/not does not change?

nodewatcher/modules/analysis/rogue_nodes/algorithm.py
```diff
@@ -0,0 +1,40 @@
...
+    three edges are present, declare that node to be rogue. If two edges are present, we perform a linear time
+    correction step. The algorithm takes at most O(EV log(V)) time.
+
+    :param graph: A graph datastructure which contains both friendly and unknown nodes.
+    :param friendly_nodes: An array of nodes that are monitored by nodewatcher and are considered friendly.
+    :return: An array of all unknown nodes with an associated probability of being a rogue node.
+    """
+
+    nx_graph = nx.Graph()
+    nx_graph.add_nodes_from([(node['i'], {'b': node['b']}) if 'b' in node else node['i'] for node in graph["v"]])
```
mitar: Don't use `"`.

nodewatcher/modules/analysis/rogue_nodes/tasks.py
```diff
@@ -0,0 +1,37 @@
+import datetime
+from django.core import mail
```
mitar, line 3: Newline before this import.

CdavM:
When you say "results", you mean the output of the detection algorithm? Meaning a bunch of nodes and the probability of each being rogue?
I think we need to store all the data because that should be the output of an algorithm. An alternative is to filter the results before the algorithm outputs them, but I prefer it the way it is now. 
But it is true that I am assuming that there is only one rogue node in these test cases and we are not sensitive at all to perturbations. 

So what should we do?

mitar:
I mean results of tests. So you have files with results and you compare then against those files, no? Do we really want to require that those results always in the future match?

nodewatcher/modules/analysis/rogue_nodes/tests.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django import test as django_test
+
+from nodewatcher.modules.analysis.rogue_nodes.tasks import rogue_node_detection_algorithm
+
+
+class RogueDetectionTest(django_test.TestCase):
+
+    def test_rogue_detection_algorithm(self):
+        print('testing rogue_node_detection_algorithm')
```
mitar: Debug print? Remove.

CdavM: What about the print statements that show exactly which test case failed? Should I remove those as well?

mitar: Oh. Hm. I though you should look into how to correctly do this as test. I think you should simply dynamically create a test suite (Google for that), instead of having one test for all files. So a test suite where you register one test for each file. And then the testing framework should display errors.

CdavM:
Yes I think that's a good idea. So the way I see it the algorithm works fine for Cloyne but may not work for other applications. And if someone wants to improve it they should make sure it's strictly better than the one we have right now.

How else would you do it without requiring that the Cloyne inputs produce the same output?

nodewatcher/modules/analysis/rogue_nodes/tests.py
CdavM: After much Googling:
By "dynamically creating a test suite": Do you mean a function that takes in a specific input and asserted results, then processes the specific input and compares it to asserted results? 

(something like "test_animal_pk" on https://docs.djangoproject.com/en/1.9/topics/testing/advanced/)

And then call this function with all the test cases? So have 8 function calls like test(e2-study-room), test(living-room), ... ?

mitar:
> Yes I think that's a good idea. So the way I see it the algorithm works fine for Cloyne but may not work for other applications. And if someone wants to improve it they should make sure it's strictly better than the one we have right now.

Yes, the detection should be strictly better, but not all numbers being output from the algorithm. Currently you require all the numbers to be the same, no?

> How else would you do it without requiring that the Cloyne inputs produce the same output?

For every test case you just check that the probability of the detection of the rogue node is inside some margin, so not exact value. So if it is x = 0.6 < 0.9 (0.9 is limit for example where algorithm says it is a rogue node), then you allow 0.5 < x 0.7 < 0.9 or something. So you allow probability to fluctuate a bit, but you do not allow that it changes the classification.

But it is OK. You do not have to worry about this. I am commenting so that you think a bit about this. We can also change this in the future when we get to more tests and break these tests.

nodewatcher/modules/analysis/rogue_nodes/tests.py
mitar: So you define a custom test suite:
- https://stackoverflow.com/questions/9261474/custom-test-suite-for-django-app
- http://stackoverflow.com/a/20870875/252025
- http://stackoverflow.com/a/23508426/252025

mitar:
So, what is the status of this? Do you think it is ready for a code review?

CdavM:
Yes, I implemented a custom test suite. Please do a code review.

CdavM:
We could also try to ensure that the numbers are in a certain range, as you proposed, but it is not within the scope of the algorithm to determine the "classification" of a rogue node. That is done by the maintainers. So we might be breaking abstraction barriers.

But more importantly, I don't foresee anyone tampering with this algorithm in the near future so I wouldn't spend too much time trying to predict how they will work on it.

nodewatcher/modules/analysis/rogue_nodes/algorithm.py
```diff
@@ -0,0 +1,39 @@
...
+    minimum spanning tree of the graph and then computes the probability of a node being rogue from
+     the number of edges that node has in a MST. The algorithm takes at most O(E log(V)) time.
+
+    :param graph: A graph datastructure which contains both friendly and unknown nodes.
+    :param friendly_nodes: An array of nodes that are monitored by nodewatcher and are considered friendly.
+    :return: An array of all unknown nodes with an associated probability of being a rogue node.
+    """
+
+    nx_graph = nx.Graph()
+    nx_graph.add_nodes_from([(node['i'], {'b': node['b']}) if 'b' in node else node['i'] for node in graph['v']])
```
mitar, line 16: I do not get it here. Nodes are or tuple of `(ID, {b: BSSID})` or just `ID`? This hurts my mental type-checking. :-)

nodewatcher/modules/analysis/rogue_nodes/test_json_files/behind-couch-living-room-results.json
```diff
@@ -0,0 +1,401 @@
...
+    {
+        "name": "7E:8F:E0:AA:C7:A0",
+        "probability_being_rogue": 0.0,
+        "ssids": [
+            null
+        ]
+    },
+    {
+        "name": "C4:6E:1F:C5:76:84",
+        "probability_being_rogue": 1,
```
mitar: Probability is not a float?

nodewatcher/modules/analysis/rogue_nodes/test_json_files/e3-study-room-SW-corner-results.json
```diff
@@ -0,0 +1,471 @@
...
+    {
+        "name": "7E:8F:E0:AA:C7:A0",
+        "probability_being_rogue": 0.0,
+        "ssids": [
+            null
+        ]
+    },
+    {
+        "name": "C4:6E:1F:C5:76:84",
+        "probability_being_rogue": 0.5,
```
mitar, line 298: So in this case it did not detect it with our 0.9 threshold?

nodewatcher/modules/analysis/rogue_nodes/test_json_files/mainprinter-results.json
```diff
@@ -0,0 +1,673 @@
...
+    {
+        "name": "62e3b05f-e0b9-4c4d-9961-ceba4a3df787",
+        "probability_being_rogue": 1,
+        "ssids": [
+            "CloyneSlow",
+            "eduroam",
+            "JSTJEsuits",
+            null,
+            "Allenoke-2.4",
+            "TP-LINK_C57684"
```
mitar: ???

You added fake SSIDs to it?

nodewatcher/modules/analysis/rogue_nodes/test_json_files/mainprinter-results.json
```diff
@@ -0,0 +1,673 @@
...
+    {
+        "name": "7E:8F:E0:AA:C7:A0",
+        "probability_being_rogue": 0.0,
+        "ssids": [
+            null
+        ]
+    },
+    {
+        "name": "C4:6E:1F:C5:76:84",
+        "probability_being_rogue": 0.5,
```
mitar: So this is the real rogue with < 0.9?

nodewatcher/modules/analysis/rogue_nodes/test_json_files/mainprinter-results.json
mitar: Is this another node? But why it looks registered/known? But probability to be rogue is 1? How can a known node (with UUID) have 1 as probability?

nodewatcher/modules/analysis/rogue_nodes/test_json_files/mainprinter-results.json
```diff
@@ -0,0 +1,673 @@
...
+            "TP-LINK_CD00",
+            "HOME-6D59-2.4",
+            "AirBears2",
+            "Stebbins",
+            "eduroam",
+            null
+        ]
+    },
+    {
+        "name": "df93387a-6180-4cde-a854-7cba043c385f",
```
mitar: You registered a node?

nodewatcher/modules/analysis/rogue_nodes/tests.py
```diff
@@ -0,0 +1,36 @@
...
+                # Run the algorithm on the file and store the results.
+                input_graph = json.load(io.open(os.path.join(path, filename), encoding='utf-8'))
+                algorithm_output = rogue_node_detection_algorithm(
+                    graph=input_graph['graph'],
+                    friendly_nodes=input_graph['friendly_nodes'],
+                )
+                # append -results to the end of the filename
+                results_filename = '{0}-results.json'.format(filename[:-5])
+                asserted_output = json.load(io.open(os.path.join(path, results_filename), encoding='utf-8'))
+                test_cases.addTest(RogueNodeTestCase('run_test', algorithm_output, asserted_output))
```
mitar: Lol. You cannot run tests when creating a suite. You should be only doing definition of tests during this stage and run them later on when tests run.

mitar:
I reviewed the pull request. Please answer questions I made. Also see changes I made. See if everything still works.

nodewatcher/modules/analysis/rogue_nodes/algorithm.py
CdavM, line 16: Yes, that's how you create nodes in nx:
http://networkx.readthedocs.io/en/latest/reference/generated/networkx.Graph.add_node.html#networkx.Graph.add_node

It has to be a tuple in case you're storing additional information.

nodewatcher/modules/analysis/rogue_nodes/test_json_files/e3-study-room-SW-corner-results.json
CdavM, line 298: Correct.

nodewatcher/modules/analysis/rogue_nodes/test_json_files/behind-couch-living-room-results.json
CdavM: This got fixed and it's now a float. Must be an outdated json file.

nodewatcher/modules/analysis/rogue_nodes/test_json_files/mainprinter-results.json
CdavM: I think this was the result of a bug. I should re-do this test case.

nodewatcher/modules/analysis/rogue_nodes/test_json_files/mainprinter-results.json
CdavM: Yes, I think so. I should re-do this test case.

nodewatcher/modules/analysis/rogue_nodes/test_json_files/mainprinter-results.json
CdavM: Bug. :rotating_light:

nodewatcher/modules/analysis/rogue_nodes/test_json_files/behind-couch-living-room-results.json
mitar: Then fix it in the JSON.

nodewatcher/modules/analysis/rogue_nodes/test_json_files/mainprinter-results.json
mitar: Yes, re-do these bad cases.

CdavM:
Hi, not all tests are running with your code. Django only reports running one test: comparing 'behind-couch-....' with its results. How do we fix this?

nodewatcher/modules/analysis/rogue_nodes/tests.py
```diff
@@ -0,0 +1,40 @@
...
+        self.assertEqual(algorithm_output, asserted_output)
+
+
+def load_tests(loader, tests, pattern):
+    test_cases = unittest.TestSuite()
+    for path, dirs, files in os.walk(os.path.join(os.path.dirname(__file__), 'test_json_files')):
+        for filename in files:
+            # Test every JSON file that does not contain "results" in the filename.
+            if os.path.splitext(filename)[1] == '.json' and 'results' not in filename:
+                test_cases.addTest(RogueNodeTestCase('run_test', os.path.join(path, filename)))
```
mitar, line 41: Can you check if this is called for all files?

CdavM, line 41: Before you modified the code, I was able to use print statements to debug. It seems that stdout is now being redirected. I also can't use pdb, it seems to skip over the whole thing. How do I verify that this is called?

mitar, line 41: Print statements in `load_tests` should work, because that is called when tests are being loaded (the same as before in your code). So does for loop work correctly?

Inside tests stdout is collected and you get it to output only if there is an error. So one simple way is to throw an exception with your message. ;-)

CdavM, line 41: OK, the for loop works correctly and finds all the test cases. test_cases is an array of 7 elements, one for every test we have. 
But run_test only runs once in total (only with the first test_filename).

mitar, line 41: Not sure. Research. Maybe the issue is that we have the same name for tests all the time. So [id](https://docs.python.org/3/library/unittest.html#unittest.TestCase.id) returns the same. You should probably override id and get it to return module name + class name + test_filename or something.

mitar, line 41: See how it is officially defined. You could also just call super and append fest_filename to the id.

CdavM, line 41: I decided it would be easier to construct a test_case with a directory and the test would go into the directory and check each entry. This also actually works.

mitar, line 41: Have you tried `id` method approach?

Are you sure it works? That tests don't now stop on first error? That you run all tests and allow some of them to fail?

nodewatcher/modules/analysis/rogue_nodes/tests.py
```diff
@@ -0,0 +1,40 @@
...
+                        graph=input_graph['graph'],
+                        friendly_nodes=input_graph['friendly_nodes'],
+                    )
+
+                    # Append "-results" to the end of the filename.
+                    results_filename = '{0}-results.json'.format(os.path.splitext(filename)[0])
+                    with io.open(os.path.join(path, results_filename), encoding='utf-8') as asserted_output_file:
+                        asserted_output = json.load(asserted_output_file)
+
+                    self.assertEqual(algorithm_output, asserted_output)
```
mitar: If any test fails now, other tests are not run.

nodewatcher/modules/analysis/rogue_nodes/tests.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def load_tests(loader, tests, pattern):
+    test_cases = unittest.TestSuite()
+    test_cases.addTest(RogueNodeTestCase('run_test', os.path.join(os.path.dirname(__file__), 'test_json_files')))
```
mitar: You do not need that then. You just create a test file with test case and it is found automatically. `load_tests` is used only when you want custom loading.

I find this approach now bad. The one before was better, but you have to learn to read documentation and how to fix it. Are you saying that making `id` unique did not work?

nodewatcher/modules/analysis/rogue_nodes/tests.py
CdavM, line 41: Hi,

the overriding `id` doesn't work. 
Tests probably do stop on the first error, but at least all of them run.

CdavM, line 41: ok. I found out that I have to override the hashing function. It works now. will push new code shortly.

CdavM:
@mitar Can you do a code review of this module?

mitar:
Perfect. Merging!

PR #28, opened 2016-08-10
Channel allocation module.

CdavM:
This PR was created to discuss implementation details of the channel allocation module.


nodewatcher/modules/analysis/channel_allocation/channel_lookup.py
```diff
@@ -0,0 +1,41 @@
...
+
+def get_available_channels(node_bssid):
+    """
+    Returns an array of available channels for that node.
+
+    :param node_bssid: BSSID of a node
+    :return: Array of available channels.
+    """
+    highest_2ghz_channel = 11
+    ch_2ghz = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
```
kostko: Channels for both protocols are already available in the following module, together with some other metadata:
- [nodewatcher.core.generator.cgm.protocols](https://github.com/wlanslovenija/nodewatcher/blob/development/nodewatcher/core/generator/cgm/protocols.py#L221)

If there are some channels missing, add them there and fix this code to use protocol information from there. The same also goes for your `interference` function in `allocation_algorithms`.

nodewatcher/modules/analysis/channel_allocation/channel_lookup.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def get_available_channels(node_bssid):
+    """
+    Returns an array of available channels for that node.
+
+    :param node_bssid: BSSID of a node
+    :return: Array of available channels.
+    """
+    highest_2ghz_channel = 11
```
kostko: Don't hard-code this, compute it from the channel list.

nodewatcher/modules/analysis/channel_allocation/channel_lookup.py
CdavM: The issue is that some of these channels are not available to the nodes in the US (specifically, 12 and 13 in the 2.4GHz band). Mitar and I think we should add a node-specific subset of available channels. But in the meantime, should we leave the hard-coded array or do something else?

nodewatcher/modules/analysis/channel_allocation/channel_lookup.py
CdavM: Fixed with 1613bd1.

nodewatcher/modules/analysis/channel_allocation/channel_lookup.py
kostko: > Mitar and I think we should add a node-specific subset of available channels.

IMO, the right way would be to add support for different regulatory domains to the `protocols` module. And then we can apply regulatory filters to get the correct channels.

mitar: Yes, but let's do this later. For now I think a TODO is good enough.

kostko:
I've fixed Travis CI in `development`, please rebase/merge.

kostko:
Something does not look right, why are the commits duplicated?

CdavM:
hmm I must have messed up the merge. will try again.

CdavM:
@mitar The fixtures are not being loaded on the Travis image, but it works locally (without the full path). Can you take a look?

mitar:
Interesting. Can you temporary try to run it in verbose mode? So add verbose flag to your .travis.yaml file and push it here.

CdavM:
@mitar, can you do a code review?

nodewatcher/modules/analysis/channel_allocation/migrations/0003_auto_20160822_0426.py
```diff
@@ -0,0 +1,20 @@
...
+        ('channel_allocation', '0002_remove_nodechannel_channel_width'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='nodechannel',
+            name='optimal_channel_interference',
+            field=models.FloatField(),
+        ),
+    ]
```
CdavM: Should I squash these three migrations files?

mitar: That is possible to be done? Than probably answer is yes. @kostko, have we been doing that?

CdavM: I can probably edit `0001_initial.py` with subsequent changes. Will try and see.

mitar: There is a tool for this: https://docs.djangoproject.com/en/1.9/ref/django-admin/#django-admin-squashmigrations

CdavM: OK. Did you review the rest of the code?

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
```diff
@@ -0,0 +1,327 @@
+import networkx as nx
+
+from nodewatcher.modules.analysis.channel_allocation import channel_lookup
```
mitar: Use relative imports for these. You can also combined it into one line.

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
```diff
@@ -0,0 +1,327 @@
...
+from nodewatcher.modules.analysis.channel_allocation import signal_processing
+
+
+def meta_algorithm(graph, known_nodes):
+    """
+    Ensures that specific algorithms are called with appropriate data structures and chooses the algorithm.
+
+    :param graph: A graph datastructure which contains both friendly and unknown nodes.
+    :param known_nodes: An array of BSSIDs monitored by nodewatcher.
+    :return: A dictionary with two k-v pairs, one for each frequency spectrum. Keys are "2.4GHz" and "5GHz".
```
mitar, line 12: But aren't there also on 5 GHZ two separate ranges of channels? Why putting them together? Why not having all together, or all separate?

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
```diff
@@ -0,0 +1,327 @@
...
+    """
+
+    highest_2ghz_channel = 11
```
mitar: Probably this should be set to 15 or something?

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
```diff
@@ -0,0 +1,327 @@
...
+            'current_interference': optimal_2ghz_graph[node]['current_interference'],
+        }
+
+    for node in optimal_5ghz_graph:
+        channel_allocations[node] = {
+            'freq': optimal_5ghz_graph[node]['freq'],
+            'width': optimal_5ghz_graph[node]['width'],
+            'interference': optimal_5ghz_graph[node]['interference'],
+            'current_interference': optimal_5ghz_graph[node]['current_interference'],
+
```
mitar: Empty line not necessary.

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
```diff
@@ -0,0 +1,327 @@
...
+        }
+
+    return channel_allocations
+
+
+def channel_to_frequency(channel):
+    ch_to_freq = {
+        '1': 2412,
```
mitar, line 101: Why not using that algorithm I linked to you?

Also, this function expects inputs as strings? It would be great if you would make sure it works with numbers as well. It is a bid strange API in this way.

mitar, line 101: (The API of the function is leaking information that you are getting your channels as strings, but this is independent from this function.)

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
```diff
@@ -0,0 +1,327 @@
...
+    all nodes have been assigned a frequency band in one specific channel width, the algorithm repeats the procedure
+    with wider channel widths. It expands the channel width in case that the interference in the wider band is not
+    1.2 stronger than the original interference. The constants are yet to be tweaked.
+
+    :param nx_graph: NX graph.
+    :param channels:  existing channel constraints.
+    :return: A dictionary of all nodes that were assigned a frequency band, along with its interference.
+    """
+
+    # Constant that needs to be tweaked.
```
mitar: Add TODO.

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
```diff
@@ -0,0 +1,327 @@
...
+            if current_interference < smallest_interference_so_far:
+                smallest_interference_so_far = current_interference
+                best_frequency_so_far = frequency
+
+    return best_frequency_so_far, smallest_interference_so_far, current_frequency_interference
+
+
+def list_neighbor_channels(nx_graph, node):
+    """
+    Goes through all neighbors and adds their signal strength to the interference it experiences..
```
mitar: Two dots at the end.

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
```diff
@@ -0,0 +1,327 @@
...
+            nx_graph[node][neighbor]['s'],
+            nx_graph[node][neighbor]['w']
+        )
+        for frequency in new_interference:
+            if frequency not in neighbor_channels:
+                neighbor_channels[frequency] = new_interference[frequency]
+            else:
+                neighbor_channels[frequency] = signal_processing.combine_power(
+                    neighbor_channels[frequency],
+                    new_interference[frequency]
```
mitar: Same.

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
```diff
@@ -0,0 +1,327 @@
...
+    degree.
+
+    :param nx_graph: NX Graph.
+    :return: List of nodes in nx_graph in decreasing order by degree.
+    """
+
+    return sorted(nx_graph, key=nx_graph.degree, reverse=True)
+
+
+def greedy_color_with_constraints(nx_graph, channels={}):
```
mitar, line 174: So we do not have optimal implementation of this? Even if it is NP hard? For small environments it could probably run all combinations, no?

nodewatcher/modules/analysis/channel_allocation/channel_lookup.py
```diff
@@ -0,0 +1,63 @@
...
+
+def get_channel(node_bssid):
+    """
+    Returns the current channel of the BSSID.
+
+    :param node_bssid: BSSID address on which we're performing the lookup.
+    :return: Channel currently assigned to that network interface.
+    """
+
+    return models.WifiInterfaceMonitor.objects.filter(bssid=node_bssid)[0].channel
```
mitar: You should use [get](https://docs.djangoproject.com/en/1.10/topics/db/queries/#retrieving-a-single-object-with-get) here for one object. (This also assures that there will be an exception if it matches more than one.)

mitar: Hm, which can potentially happen. I can imagine one registering a node with one hardware and then moving that same hardware to another node. So we have the same MAC twice. But probably they will not be part of the same survey twice because they will not be in the air at the same time. @kostko, what do you think? I think get is enough here and exceptions we do not have to worry about probably. We can address this later if we discover that they are common.

nodewatcher/modules/analysis/channel_allocation/channel_lookup.py
```diff
@@ -0,0 +1,63 @@
+from nodewatcher.core.monitor import models
+
+
+# TODO: Turn hard-coded arrays into a node-specific (stemming from regulatory practices) property.
+ch_2ghz = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
+ch_5ghz = [36, 38, 40, 42, 44, 46, 48, 50, 52, 54, 56, 58, 60, 62, 64, 100, 102, 104, 106, 108, 110, 112, 114, 116,
+           118, 120, 122, 124, 126, 128, 132, 134, 136, 138, 140, 142, 144, 149, 151, 153, 155, 157, 159, 161, 165]
+highest_2ghz_channel = max(ch_2ghz)
```
mitar, line 8: Can you make this then be used in the other file as well? Instead of having a constant there?

nodewatcher/modules/analysis/channel_allocation/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+from nodewatcher.core.monitor import models as monitor_models
+
+
+class NodeChannel(models.Model):
+
```
mitar: Unnecessary line?

nodewatcher/modules/analysis/channel_allocation/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+from nodewatcher.core.monitor import models as monitor_models
```
mitar, line 3: Empty line before this import.

nodewatcher/modules/analysis/channel_allocation/tasks.py
```diff
@@ -0,0 +1,40 @@
...
+    interface_dict = allocation_algorithms.meta_algorithm(extracted_graph['graph'], extracted_graph['known_nodes'])
+
+    for interface in interface_dict:
+        n, created = models.NodeChannel.objects.get_or_create(
+            interface=wifi_models.WifiInterfaceMonitor.objects.get(bssid=interface),
+            defaults={
+                'optimal_start_frequency': interface_dict[interface]['freq'],
+                'optimal_channel_width': interface_dict[interface]['width'],
+                'optimal_channel_interference': interface_dict[interface]['interference'],
+            }
```
mitar: Comma at the end.

nodewatcher/settings.py
```diff
@@ -250,6 +250,7 @@ def user_url(user):
 
     # Modules.
     #'nodewatcher.modules.analysis.rogue_nodes',
+    'nodewatcher.modules.analysis.channel_allocation',
```
mitar: I think we should find a way to make test be skipped if its app it is not added to settings.

mitar:
Otherwise it looks pretty good. Just minor things. I do not know about the algorithm, though. I hope that it works. ;-)

nodewatcher/modules/analysis/channel_allocation/channel_lookup.py
kostko: Yes I think using get here is ok. BTW, why do you repeat the exact same code all over below? I don't like this, the queryset should be done only in one place, perhaps a function like `get_interface_for_bssid(bssid)` or something.

nodewatcher/modules/analysis/channel_allocation/migrations/0003_auto_20160822_0426.py
kostko: I have squashed migrations when I also squashed commits (but not using Django's squashmigrations). But I think that a single PR should not introduce multiple migrations for a single app as this is really bad for the time it takes to apply the migrations.

CdavM: I have squashed the migrations.

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
CdavM: Mostly because of the consistency with our data collection. We use the same network interface for both ranges on 5GHz, and a different one for 2.4GHz. How should we clarify this?

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
CdavM: This doesn't expect a string as input, it casts the type to a string should it be a number. But I can add some documentation for it.

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
CdavM: What does "Same" mean? Same as what?

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
CdavM: No, this is our only algorithm for now. We could probably implement a thorough search as well.

nodewatcher/modules/analysis/channel_allocation/channel_lookup.py
CdavM: Done

nodewatcher/modules/analysis/channel_allocation/models.py
CdavM: Removed.

nodewatcher/settings.py
CdavM: We are no longer installing this app by default and we're skipping the test if it's not installed. Although there are still warnings in the log because the fixtures are not being loaded properly (if the app is not installed).

nodewatcher/modules/analysis/channel_allocation/channel_lookup.py
CdavM: Good idea. 

I have created an additional helper so that we have the query in just one location.

CdavM:
@mitar I made all the changes requested. One more code review :-) ?

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
mitar: As comment above. Comma is missing. Just nitpicking.

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
mitar: Could you do it in the followup pull request please?

mitar:
LGTM.

CdavM:
OK so let's merge it :-)

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
mitar: BTW, this still holds. ;-) You didn't answer. ;-)

mitar:
Great! Great job!!!

nodewatcher/modules/analysis/channel_allocation/allocation_algorithms.py
CdavM: Yep. But since we have 26 different start frequencies in the 5ghz band, this is 26^10 combinations for only 10 nodes. I was thinking of something like this:
while()
  generate a random allocation
  if it's better than the best so far:
    this is the current best
return current best.

What do you think?

mitar: This is not a search algorithm. :-) It looks like genetic algorithm, only that previous generation and next generation do not have anything in common. :-)

I think the best would be to use some general optimization tool and then express the problem in that language of the tool and then use one of their algorithms for optimization.

mitar: Is there a library which is already doing graph coloring?

CdavM: There is NetworkX (similar to our current implementation), but I have found no libraries that do exact graph coloring.

mitar: This one has some algorithms: https://pypi.python.org/pypi/pyclustering

mitar: But yes, it really seems most are doing greedy.

PR #29, opened 2016-08-12
Added more channels.

CdavM:
I added all the channels from https://en.wikipedia.org/wiki/List_of_WLAN_channels#5.C2.A0GHz_.28802.11a.2Fh.2Fj.2Fn.2Fac.29.5B17.5D
to the list of 5GHz channels.


CdavM:
@mitar @kostko Can you take a look?

PR #30, opened 2016-09-11
Update map.html

mitjajez:
may fix "same id" problem - h1 and map container have id 'map'


mitar:
Oh, I would prefer if you would change the ID of the div into which we are putting map and leave title as it is.

mitar:
(You can just update this pull request with a new commit.)

PR #31, opened 2016-09-11
Update map.html

mitjajez:
Map renders into same id as title

<h1 id="map" class="heading ">Map</h1>

<div id="map" class="leaflet-container-default"></div>

this should fix it


mitar:
Why opening a new pull request? Why not updating the old one?

And no, this is also not good enough. Because you have also to change the JavaScript code to point it to this new coordinate.

I think you should clone the code locally and edit there in your branch, instead of using GitHub to do so.

Also, you should probably test the code before making a pull request and saying that this is a fix. The whole point of a contribution is that you make the effort to test it so that we can then distribute the work. Just editing the file without testing means that then I have to test it, so how much is then this really helpful?

PR #32, opened 2016-09-18
description: Display node's URL.

jaka:


kostko (changes requested):
Nice work!

But, your code is in the wrong module. Since the description configuration is provided in `nodewatcher.modules.administration.description`, the URL partial should be placed there.

kostko (changes requested):
Also, the commit author's e-mail address doesn't seem to be linked to your GitHub account. Can you also update this?

kostko (changes requested):
And another thing, please update your commit message to include the module name prefix, as the change only applies to one module.

Something like:

```
description: Display node's URL.
```

jaka:
OK, I'll fix this. I just added code to the place where I found similar code.

PR #34, opened 2016-10-13
Switch configuration

kostko:
Changed:
- Two new configuration registry items, `core.switch` and `core.switch.vlan` with their implementations `SwitchConfig` and `VLANConfig`.
- A switch defaults class, which knows how to automatically configure switches based on default device VLAN presets, when editing configuration via the forms UI.

Missing:
- [x] Update the OpenWrt CGMs so that they know how to generate switch and interface configurations from the platform-independent configuration.
- [x] Configuration migrations from old switched ethernet port names (eg. `wan0` to `sw0.vlan1`).
- [x] Allow multiple presets to be specified in device descriptors (besides "default"), rename `SwitchedEthernetPort` to `SwitchVLANPreset`.
- [x] Update _wlan slovenija_ configuration defaults to support new switched ports.
- [x] Update core schema documentation for new switch-related registry items.


mitar:
cc @CdavM

PR #35, opened 2017-04-14
Add profile for Ubiquiti Edgerouter X

robimarko:
Initial ERX support


mitar:
Please fix so that tests do not fail.

robimarko:
Finally got it to build without errors

mitar:
While this is passing tests, can you still make sure that code style matches other files? Pay close attention to newlines and code structure.

robimarko:
I compared this with one of your templates and I think that now it matches

mitar (approved):

nodewatcher/modules/devices/ubnt/edgerouter.py
```diff
@@ -0,0 +1,48 @@
...
+        'lede': {
+            'name': 'ubnt-er-x',
+            'files': [
+                '*-mt7621-generic-ubnt-er-x-squashfs-factory.bin',
+                '*-mt7621-generic-ubnt-er-x-squashfs-sysupgrade.bin',
+            ]
+        }
+    }
+
+#Register Edgerouter X device
```
kostko: Space after `#`, dot at the end.

nodewatcher/modules/devices/ubnt/edgerouter.py
```diff
@@ -0,0 +1,48 @@
...
+
+from nodewatcher.core.generator.cgm import base as cgm_base, protocols as cgm_protocols, devices as cgm_devices
+
+
+class UBNTEdgerouterX(cgm_devices.DeviceBase):
+    """
+    UBNT EdgerouterX device descriptor.
+    """
+
+    identifier = 'ub-er-x'
```
kostko: Change identifier to `ub-edgerouter-x`.

nodewatcher/modules/devices/ubnt/edgerouter.py
```diff
@@ -0,0 +1,48 @@
...
+            ports=5,
+            cpu_port=0,
+            vlans=16,
+            configurable=True,
+            presets=[
+                cgm_devices.SwitchPreset('default', _("Default VLAN configuration"), vlans=[
+                    cgm_devices.SwitchVLANPreset(
+                        'lan0', "Lan0",
+                        vlan=1,
+                        ports=[0, 1, 2, 3, 4],
```
kostko: Did you check that this configuration is correct?

kostko (changes requested):
Some minor comments. It looks like the device descriptor is missing port mappings, see `port_map` in other descriptors. You have to configure this based on how the ports are called in LEDE.

Also, please squash all your commits into a single commit as this is a lot of changes to a single new file.

nodewatcher/modules/devices/ubnt/edgerouter.py
robimarko: From what I know and from Lede source it should be correct.

robimarko:
I have tried to make changes to comply with your requests

nodewatcher/modules/devices/ubnt/edgerouter.py
```diff
@@ -0,0 +1,54 @@
...
+                ])
+            ]
+        )
+    ]
+    ports = [
+        cgm_devices.EthernetPort('wan0', "Wan0"),
+    ]
+    port_map = {
+        'openwrt': {
+            'sw0': cgm_devices.SwitchPortMap('switch0', vlans='eth0'),
```
kostko: Port mapping seems incorrect. You have set static `eth0` for all VLANs and also `wan0`. If the switch actually supports VLANs, the mapping should be something like `eth0.{vlan}`. But you need to set this based on how the switch is configured in LEDE, please check that (maybe link to the section in LEDE source where defaults for this device are configured).

But especially you can't have both the switch and `wan0` on the same interface (e.g. `eth0`). If the WAN port is actually part of the switch, it needs to be configured as a switch VLAN preset, similar to the LAN port (with the correct port definitions of course).

robimarko: Well that be beyond my understanding but I will try to figure it out.
Here is switch configuration in Lede source
https://github.com/lede-project/source/blob/master/target/linux/ramips/base-files/etc/board.d/02_network

ucidef_add_switch "switch0" \
			"1:lan" "2:lan" "3:lan" "4:lan" "0:wan" "6@eth0"

kostko: To me, this reads as follows:
* All ports (LAN and WAN) are connected via the switch (so your configuration above is wrong, you should not define the WAN port separately).
* VLANs are attached to `eth0`, so they will be `eth0.{vlan}`.
* Ports `1, 2, 3, 4` are assigned to the LAN VLAN.
* Port `0` is assigned to the WAN VLAN.
* Port `6` is connected to the CPU (so it must be in all VLANs).

I am not sure, so you should really check this on an actual device (e.g. with a default LEDE install).

robimarko: Ok,@valentt has one flashed with LEDE so I will check it out.

robimarko:
Here is output from swconfig on Lede
[ERX SWCONFIG.txt](https://github.com/wlanslovenija/nodewatcher/files/995873/ERX.SWCONFIG.txt)

kostko:
Can you also paste the output of `uci show network`?

robimarko:
Sure,here it is
[ERX UCI.txt](https://github.com/wlanslovenija/nodewatcher/files/995872/ERX.UCI.txt)

robimarko:
I have finally got around to fixing switch configuration.
Now it should be correct as per LEDE

Had to fix syntax a bit

kostko:
Looks good (note that we haven't yet built LEDE for `ramips_mt7621`), it would just be nice if you could squash all your commits into a single one and change the commit message to `devices: Add profile for Ubiquiti Edgerouter X.` Do you know how to squash commits?

robimarko:
Unfortunatelly not.

kostko:
Please first rebase on `development` as there have been some updates to that branch (802.11ac support).

robimarko:
Should be done.

kostko:
Ok, so you merged instead of rebasing, which introduced yet more commits :-)

Are you using the command-line Git client? If so, do the following to correctly rebase your commits and squash them.

First get the latest changes and rebase (instead of merge):
```
git fetch origin
git rebase origin/development
```

Then squash your last 10 commits by running:
```
git rebase -i HEAD~10
```

This will open an editor that will enable you to interactively update the commits, sorted from oldest to newest.  For the first one, keep the `pick` command. For all the next ones change `pick` to `squash`. Save and exit the editor.

Then, the editor will stop for squashing commits. Edit the commit message to say `devices: Add profile for Ubiquiti Edgerouter X.` (e.g., remove all the other lines). Save and exit the editor.

This should now give you a single commit in your history containing all your changes (check this with `git log`). Once you are sure that everything is correct, run `git push -f` to overwrite your PR branch.

I can do all of this for you, but I would like you to learn how to do this yourself, so it will be easier next time.

robimarko:
I am trying but I can find commits for Edgerouter X,only UAP

kostko:
Be sure that you are on the correct branch before doing anything. To switch to this branch do `git checkout patch-1` before all other commands.

robimarko:
Now it should be done

kostko:
Nope, something does not seem right? You seem to have combined the two different branches into one? Just use `git rebase -i HEAD~10` to remove the commits that should not be there (remove the lines with invalid commits). But before doing that, make sure you are on the correct branch.

kostko:
Also it seems that you don't have the correct e-mail address configured in `git`?

robimarko:
I dont get how could I combine two branches?
I used patch-1 branch and then squashed commits into one.
I used same email as in my github account

mitar:
I think the problem is that @robimarko is working on `development` branch for #36.

@robimarko, please **do not** now just close this or the other pull request and open new ones, but fix them instead. Learn to do that.

robimarko:
Proper PR done.
Check PR #39

PR #37, opened 2017-05-13
cgm: Add preliminary 802.11ac protocol support

kostko:


robimarko (approved):
Should work from what I see

PR #38, opened 2017-05-14
Add Unifi UAP and UAP AC Lite device descriptors

robimarko:
Pull request to add Unifi UAP and newer UAP AC Lite device descriptors.
This time properly made in separate branch.

### DIFF
--- a/nodewatcher/modules/devices/ubnt/__init__.py
+++ b/nodewatcher/modules/devices/ubnt/__init__.py
@@ -1,2 +1,2 @@
 # UBNT device modules.
-from . import nano, bullet, rocket, pico, powerbeam, edgerouter
+from . import nano, bullet, rocket, pico, powerbeam, edgerouter, unifi
--- a/nodewatcher/modules/devices/ubnt/unifi.py
+++ b/nodewatcher/modules/devices/ubnt/unifi.py
@@ -1,0 +1,157 @@
+from django.utils.translation import ugettext_lazy as _
+
+from nodewatcher.core.generator.cgm import base as cgm_base, protocols as cgm_protocols, devices as cgm_devices
+
+
+class UBNTAPACLite(cgm_devices.DeviceBase):
+    """
+    Ubiquiti AP AC Lite device descriptor.
+    """
+
+    identifier = 'ub-uap-ac-lite'
+    name = "Unifi AP AC Lite"
+    manufacturer = "Ubiquiti"
+    url = 'http://www.ubnt.com/'
+    architecture = 'ar71xx'
+    radios = [
+        cgm_devices.IntegratedRadio('wifi0', _("Integrated wireless radio (5 GHz)"), [
+            cgm_protocols.IEEE80211AC(
+                cgm_protocols.IEEE80211AC.SHORT_GI_20,
+                cgm_protocols.IEEE80211AC.SHORT_GI_40,
+                cgm_protocols.IEEE80211AC.RX_STBC1,
+                cgm_protocols.IEEE80211AC.DSSS_CCK_40,
+            )
+        ], [
+            cgm_devices.AntennaConnector('a1', "Antenna1")
+        ], [
+            cgm_devices.DeviceRadio.MultipleSSID,
+        ]),
+        cgm_devices.IntegratedRadio('wifi1', _("Integrated wireless radio (2.4 GHz)"), [
+            cgm_protocols.IEEE80211BGN(
+                cgm_protocols.IEEE80211BGN.SHORT_GI_20,
+                cgm_protocols.IEEE80211BGN.SHORT_GI_40,
+                cgm_protocols.IEEE80211BGN.RX_STBC1,
+                cgm_protocols.IEEE80211BGN.DSSS_CCK_40,
+            )
+        ], [
+            cgm_devices.AntennaConnector('a2', "Antenna2")
+        ], [
+            cgm_devices.DeviceRadio.MultipleSSID,
+        ])
+    ]
+    switches = []
+    ports = [
+        cgm_devices.EthernetPort('lan0', "Lan0")
+    ]
+    antennas = [
+        # TODO: This information is probably not correct
+        cgm_devices.InternalAntenna(
+            identifier='a1',
+            polarization='dual',
+            angle_horizontal=360,
+            angle_vertical=120,
+            gain=3,
+        ),
+        cgm_devices.InternalAntenna(
+            identifier='a2',
+            polarization='dual',
+            angle_horizontal=360,
+            angle_vertical=120,
+            gain=3,
+        )
+    ]
+    port_map = {
+        'openwrt': {
+            'wifi0': 'radio0',
+            'wifi1': 'radio1',
+            'lan0': 'eth0',
+        }
+    }
+    drivers = {
+        'openwrt': {
+            'wifi0': 'mac80211',
+            'wifi1': 'mac80211',
+        }
+    }
+    profiles = {
+        'lede': {
+            'name': 'ubnt-unifiac-lite',
+            'files': [
+                '*-ar71xx-generic-ubnt-unifiac-lite-squashfs-sysupgrade.bin',
+            ]
+        }
+    }
+
+
+class UBNTUap(cgm_devices.DeviceBase):
+    """
+    UBNT Unifi UAP device descriptor.
+    """
+
+    identifier = 'ub-uap'
+    name = "Unifi UAP"
+    manufacturer = "Ubiquiti"
+    url = 'http://www.ubnt.com/'
+    architecture = 'ar71xx'
+    radios = [
+        cgm_devices.IntegratedRadio('wifi1', _("Integrated wireless radio (2.4 GHz)"), [
+            cgm_protocols.IEEE80211BGN(
+                cgm_protocols.IEEE80211BGN.SHORT_GI_20,
+                cgm_protocols.IEEE80211BGN.SHORT_GI_40,
+                cgm_protocols.IEEE80211BGN.RX_STBC1,
+                cgm_protocols.IEEE80211BGN.DSSS_CCK_40,
+            )
+        ], [
+            cgm_devices.AntennaConnector('a1', "Antenna1")
+        ], [
+            cgm_devices.DeviceRadio.MultipleSSID,
+        ])
+    ]
+    switches = []
+    ports = [
+        cgm_devices.EthernetPort('lan0', "Lan0")
+    ]
+    antennas = [
+        cgm_devices.InternalAntenna(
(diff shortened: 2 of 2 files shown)


nodewatcher/modules/devices/ubnt/__init__.py
```diff
@@ -1,2 +1,2 @@
 # UBNT device modules.
-from . import nano, bullet, rocket, pico, powerbeam
+from . import nano, bullet, rocket, pico, powerbeam ,unifi
```
mitar: Can you be a bit more precise about syntax? See wrong comma place?

nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -0,0 +1,145 @@
+from django.utils.translation import ugettext_lazy as _
+
+from nodewatcher.core.generator.cgm import base as cgm_base, protocols as cgm_protocols, devices as cgm_devices
+
+
+class UBNTUapACLite(cgm_devices.DeviceBase):
```
mitar: Just name `UbiquitiAPACLite`.

nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    UBNT UAP AC Lite device descriptor.
```
mitar: Write `Ubiquiti AP AC Lite device descriptor`.

nodewatcher/modules/devices/ubnt/__init__.py
robimarko: Yes,you are right.
Thought that I had it properly made.

nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+
+    identifier = 'ub-uap-ac-lite'
```
mitar: `ub-ap-ac-lite`.

nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -0,0 +1,145 @@
...
+    radios = [
+        cgm_devices.IntegratedRadio('wifi0', _("Integrated wireless radio (5 GHz)"), [
+            cgm_protocols.IEEE80211AC(
+                cgm_protocols.IEEE80211AC.SHORT_GI_20,
+                cgm_protocols.IEEE80211AC.SHORT_GI_40,
+                cgm_protocols.IEEE80211AC.RX_STBC1,
+                cgm_protocols.IEEE80211AC.DSSS_CCK_40,
+            )
+        ], [
+            cgm_devices.AntennaConnector('a1', "Antenna1")
```
mitar, line 25: Does it have only one antenna? Didn't you say that this is difference between models? How many antennas they have?

nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -0,0 +1,145 @@
...
+        ]),
+        cgm_devices.IntegratedRadio('wifi1', _("Integrated wireless radio (2.4 GHz)"), [
+            cgm_protocols.IEEE80211BGN(
+                cgm_protocols.IEEE80211BGN.SHORT_GI_20,
+                cgm_protocols.IEEE80211BGN.SHORT_GI_40,
+                cgm_protocols.IEEE80211BGN.RX_STBC1,
+                cgm_protocols.IEEE80211BGN.DSSS_CCK_40,
+            )
+        ], [
+            cgm_devices.AntennaConnector('a2', "Antenna0")
```
mitar: Why is this `a2` but `Antenna0`? Can you match numbers?

nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -0,0 +1,145 @@
...
+        ], [
+            cgm_devices.DeviceRadio.MultipleSSID,
+        ])
+    ]
+    switches = []
+    ports = [
+        cgm_devices.EthernetPort('lan0', "Lan0")
+    ]
+    antennas = [
+        # TODO: This information is probably not correct
```
mitar, line 47: Can you research how many antennas it has?

nodewatcher/modules/devices/ubnt/unifi.py
robimarko: No, that is AC model.
It has antenna for 2.4Ghz and another one for 5Ghz

mitar: So what is difference between? UniFi AP Lite and UniFi AP Pro?

nodewatcher/modules/devices/ubnt/unifi.py
mitar: Or, are we matching OpenWrt here? @kostko?

nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+
+class UBNTUap(cgm_devices.DeviceBase):
+    """
+    UBNT Unifi UAP device descriptor.
+    """
+
+    identifier = 'ub-uap'
+    name = "Unifi UAP"
```
mitar, line 92: Which one is this?

nodewatcher/modules/devices/ubnt/unifi.py
robimarko: I have already stated that.
Difference is in the construction of the antennas,Lite has 2x2 MIMO for 2.4Ghz and another 2x2 MIMO for 5GHz while Pro has 3x3 MIMO for 2.4GHz and another one 3x3 MIMO for 5Ghz

mitar:
Can you really please name those descriptors well? I still do not understand what they are for? Which exactly models?

nodewatcher/modules/devices/ubnt/unifi.py
robimarko: The original Unifi UAP and UAP-LR
https://www.ubnt.com/unifi/unifi-ap/

robimarko:
Look, I have named them as per product name.
I don't know hot to name the differently.

The one that you requested name to be changed to Ubiquiti AP AC Lite device descriptor is this one
https://www.ubnt.com/unifi/unifi-ap-ac-lite/

And the second one name UAP is this, actually for UAP and UAP-LR models.
https://www.ubnt.com/unifi/unifi-ap/

mitar:
I know of:
* `UniFi AP`
* `UniFi AP LR`
* `UniFi AP Pro`
* `UniFi AP AC HD`
* `UniFi AP AC LR`
* `UniFi AP AC Lite`

So the one in this pull request are `UniFi AP` and `UniFi AP AC Lite`?

robimarko:
The first one is AP AC Lite and the second one is UniFi AP and Unifi AP LR since they use the exactly same image and only differ in 2x2 vs 3x3 antennas on LR model.

mitar:
So can you make three classes? For all those three devices? You can also extend `Unifi AP LR` from `UniFi AP` class and just override the antennas, if it is different?

We want to be more precise than just wrapping filenames from OpenWrt. We want to know for each model/device, even if they share the same binary.

mitar:
For antennas, you can check in their datasheets:

https://dl.ubnt.com/datasheets/unifi/UniFi_AC_APs_DS.pdf

For `UniFi AP AC Lite` it says: "(2) Dual-Band Antennas, 3 dBi Each"

So it has two antennas, and the same antenna is used for both 2.4 GHz and 5 GHZ, and it is 3 dbi.

Pro has 3 dual-band antennas.

robimarko:
There is no need for additional AP LR class since according to datasheet AP and AP LR have the same antennas.
Integrated 3 dBi Omni (Supports 2x2 MIMO with Spatial Diversity) as per UBNT datasheet.
It looks only the output power is different

mitar:
Polarization for those antennas seems to be `dual`. `angle_vertical` seems closer to 120 degrees, [based on this diagram](https://community.ubnt.com/t5/UniFi-Wireless/Antenna-radiation-pattern-for-UniFi-AP-diagramm-inside/td-p/772468).

mitar:
> There is no need for additional AP LR class since according to datasheet AP and AP LR have the same antennas.

So just extend the class and change the identifier and description and this is it then.

We want to have precise information and know which device is which so that we could in the future compute coverage automatically and things like that.

robimarko:
I have extended the class and added AP LR descriptor.
Also, I have modified gain to be correct and vertical angle like you pointed out.

robimarko:
Just figured out that I had wrong device profile for making defined.
Now it is per LEDE profile.
@kostko @mitar Any chance of merging?

robimarko:
I have squashed previous commits into one.

kostko:
Please rebase on development and fix conflicts.

robimarko:
Will do.

robimarko:
@kostko Rebased and conflict should be fixed

kostko:
Please add the `devices: ` prefix to the commit message.

nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -0,0 +1,157 @@
...
+    port_map = {
+        'openwrt': {
+            'wifi0': 'radio0',
+            'wifi1': 'radio1',
+            'lan0': 'eth0',
+        }
+    }
+    drivers = {
+        'openwrt': {
+            'wifi0': 'ath10k',
```
kostko: Is this driver specification correct? Isn't `ath10k` also `mac80211`-based?

nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -0,0 +1,157 @@
+from django.utils.translation import ugettext_lazy as _
+
+from nodewatcher.core.generator.cgm import base as cgm_base, protocols as cgm_protocols, devices as cgm_devices
+
+
+class UbiquitiAPACLite(cgm_devices.DeviceBase):
```
kostko: This does not seem to follow the naming convention, change `Ubiquiti` to `UBNT`.

nodewatcher/modules/devices/ubnt/unifi.py
robimarko: Yes,you are right.
After another look in LEDE wireless configuration it uses mac80211 for both radios.
Will fix that now.

nodewatcher/modules/devices/ubnt/unifi.py
robimarko: Off course,will do that now

robimarko:
@kostko Changes made as you requested and prefix added.

kostko (approved):

PR #39, opened 2017-05-14
Add Edgerouter X device descriptor

robimarko:
Pull request to add Ubiquti Edgerouter X device descriptor.
This time properly made in separate branch.

### DIFF
--- a/nodewatcher/modules/devices/ubnt/__init__.py
+++ b/nodewatcher/modules/devices/ubnt/__init__.py
@@ -1,2 +1,2 @@
 # UBNT device modules.
-from . import nano, bullet, rocket, pico, powerbeam
+from . import nano, bullet, rocket, pico, powerbeam, edgerouter
--- a/nodewatcher/modules/devices/ubnt/edgerouter.py
+++ b/nodewatcher/modules/devices/ubnt/edgerouter.py
@@ -1,0 +1,59 @@
+from django.utils.translation import ugettext_lazy as _
+
+from nodewatcher.core.generator.cgm import base as cgm_base, protocols as cgm_protocols, devices as cgm_devices
+
+
+class UBNTEdgerouterX(cgm_devices.DeviceBase):
+    """
+    UBNT EdgerouterX device descriptor.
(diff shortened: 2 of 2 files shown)


robimarko:
@kostko Does Nodewatcher include CPU port in ports variable?

kostko:
> @kostko Does Nodewatcher include CPU port in ports variable?

No, the `ports` must include the CPU port identifier, otherwise this is an error. This is why you are getting the following exception:
```
ImproperlyConfigured: Switch descriptor 'sw0' refers to an invalid CPU port '6'!
```

You configured `ports=5` on the switch, but the switch as you defined it actually has 7 ports (e.g., 0, 1, 2, 3, 4, 5, 6), not just 5. So it should be `ports=7`.

robimarko:
I get I know.
I did not know that it has to include CPU port as well.

According to Lede source and actuall running Lede on ERX port 5 is not being used.
And port 6 is CPU.

Ok,will edit config to 7 ports.
Have you checked out Unifi PR#38?

kostko:
> According to Lede source and actuall running Lede on ERX port 5 is not being used.

You do not need to specify sequential ports, instead of `ports=7` you can say `ports=[0, 1, 2, 3, 4, 6]` and then port 5 will be invalid.

robimarko:
Thanks,that makes sense.
Also thanks for patience

robimarko:
Should be fixed now

robimarko:
Just figured out that I had wrong device profile for making defined.
Now it is per LEDE profile.
@kostko @mitar Any chance of merging?

robimarko:
I have squashed previous commits into one.

robimarko:
@kostko Is ramips_mt7621 builder ready?

kostko:
> @kostko Is ramips_mt7621 builder ready?

Yes.

kostko (approved):
Looks good, I think we can merge and you can fix things if somethihng shows itself to be wrong.

robimarko:
Sure, that would be great.

PR #42, opened 2017-05-21
Django 1.10

kostko:


mitar:
;-)

Sorry, didn't have time.

Dockerfile
```diff
@@ -18,8 +18,8 @@ RUN pip install --upgrade --force-reinstall pip six requests && \
     cat /code/requirements-readthedocs.txt /code/requirements.txt | xargs -n 1 sh -c 'CPLUS_INCLUDE_PATH=/usr/include/gdal C_INCLUDE_PATH=/usr/include/gdal pip install $0 || exit 255'
 
 # Remove unneeded build-time dependencies
-RUN apt-get purge python-dev build-essential -y --force-yes && \
-    apt-get autoremove -y --force-yes && \
```
mitar, line 22: I am still leaving `--force-yes`, despite the warning. :-)

nodewatcher/modules/events/sinks/db_sink/templates/events/list.html
```diff
@@ -1,29 +1,30 @@
...
+    {% add_data "js_data" "api/js/base.js" %}
+    {% add_data "js_data" "api/js/registry.js" %}
     {% add_data "js_data" "events/list/js/code.js" %}
 
     {% trans "Events" as title %}
 {% endcontextblock %}
 
 {% block altlinks %}
-    {% url "api:api_dispatch_list" "v1" "event" as api_url %}
+    {% url "apiv2:serializednodeevent-list" as api_url %}
```
mitar, line 20: Do we still have API v1? Maybe it would be more reasonable to replace `api` with `apiv1` and then have `api` for `apiv2`.

nodewatcher/modules/frontend/api/templates/rest_framework/base.html
```diff
@@ -1,21 +1,230 @@
...
 {% block head %}
     {% add_data "css_data" "rest_framework/css/prettify.css" %}
     {% add_data "css_data" "rest_framework/css/default.css" %}
 {% endblock %}
 
-{% block posthead %}{% endblock %}
-{% block navbar %}{% endblock %}
-{% block bodystart %}{% endblock %}
-{% block bodyend %}{% endblock %}
-{% block postbody %}{% endblock %}
```
mitar, line 16: Why removing these?

nodewatcher/modules/frontend/api/templates/rest_framework/base.html
```diff
@@ -1,21 +1,230 @@
-{% extends "rest_framework/base.html" %}
```
mitar, line 1: And this?

nodewatcher/modules/events/sinks/db_sink/templates/events/list.html
kostko: Only datastream is left in the v1 API. Everything else has been removed.

nodewatcher/modules/frontend/api/templates/rest_framework/base.html
kostko: From my understanding this required a fork of DRF. I don't like that, so I changed it in a way that the fork was not required and bumping versions is easier. Unfortunately it seems like they didn't agree with your proposed changes which would make it more modular ;-)

nodewatcher/modules/frontend/api/templates/rest_framework/base.html
kostko: See my comment below.

nodewatcher/modules/frontend/api/templates/rest_framework/base.html
mitar: So then we are now simply overriding their template fully? Hmm. Should we then extend from our base template?

nodewatcher/modules/events/sinks/db_sink/templates/events/list.html
mitar: Yes, so should we change it to `api` and the old one to `apiv1` for now?

I would then migrate also datastream (after we discuss how to do it, find me on chat), and then I would release 3.0 finally and merge `development` branch into `master` and start having releases.

PR #44, opened 2017-05-21
devices: Remove factory binary and fix sysupgrade name in Edgerouter X

robimarko:
Fixing my overlook in device descriptor for Edgerouter X.
Factory binary is not generated by default and therefore while building error occures.

Fixed by removing it from profiles in device descriptor

### DIFF
--- a/nodewatcher/modules/devices/ubnt/edgerouter.py
+++ b/nodewatcher/modules/devices/ubnt/edgerouter.py
@@ -49,8 +49,7 @@ class UBNTEdgerouterX(cgm_devices.DeviceBase):
         'lede': {
             'name': 'ubnt-erx',
             'files': [
-                '*-mt7621-generic-ubnt-erx-squashfs-factory.bin',
-                '*-mt7621-generic-ubnt-erx-squashfs-sysupgrade.bin',
+                '*-ramips-mt7621-ubnt-erx-squashfs-sysupgrade.tar',
             ]
         }
     }


nodewatcher/modules/devices/ubnt/edgerouter.py
```diff
@@ -50,7 +50,6 @@ class UBNTEdgerouterX(cgm_devices.DeviceBase):
             'name': 'ubnt-erx',
             'files': [
                 '*-mt7621-generic-ubnt-erx-squashfs-factory.bin',
-                '*-mt7621-generic-ubnt-erx-squashfs-sysupgrade.bin',
```
kostko: You removed sysupgrade not factory? This is probably not correct?

robimarko: Yes,I cant belive how stupid I am today
Thanks for catching it.
I will look better from now on

robimarko: Fixed it now

robimarko:
Fixed my big mistake of removing sysupgrade instead of factory

kostko:
Please update your commit to have the same message as the pull request title (also in the other PR).

robimarko:
@kostko Updated in both PRs

robimarko:
I have figured out that like in PR #45 output image name was wrong so it is also fixed.
PR name and commit message was edited to include that

mitar:
Could we make it so that factory image is generated? How is the initial flashing process anyway? Is factory image needed? If so, we should make it so that it is available.

robimarko:
Well, It could be made but factory image is not used in flashing.
Kernel image is needed, but I have already compiled LEDE kernel month or two ago and that one can be used.

Inital flashing process is shortly like this:
1.You gotta connect to serial
2.You load kernel image to RAM over built-in TFTP client
3.After kernel is loaded you use sysupgrade -V to flash sysupgrade image.

mitar:
OK, then we have to provide that kernel image through nodewatcher. How else can a new person flash the device?

mitar:
So we should provide two files: kernel image, and sysupgrade. Can you look into this?

robimarko:
Yes, we have to provide kernel and sysupgrade images.
But you don't have to generate kernel image as that requires modifying LEDE make profile, you can simply upload my prebuilt kernel image and link to it so it can be downloaded.

mitar:
> But you don't have to generate kernel image as that requires modifying LEDE make profile, you can simply upload my prebuilt kernel image and link to it so it can be downloaded.

Ehm, I do not like this. Let's make it proper.

robimarko:
Well, then make profile requires modifying.
@kostko Can certainly tell us but I think that then whole builder needs to be rebuilt.
Also I don't see how to easily modify make profile just for Edgerouter X.

kostko:
I think we can merge this and add the other part later. Is this similar to Mikrotik images where you first need to load a special ram-only image, which needs to be built in a different manner? Can you link to some documentation describing this?

And rebase on development.

robimarko:
No,it is not build specially,in menu makeconfig you just select ramdisk to be built also and then kernel image as well as factory tar which is not used is generated.

Here is good article on installing OpenWRT on Edgerouter X,but the same applies to LEDE.
Only difference is different architecture name in make profile that is outdated in OpenWRT.
http://sector5d.org/openwrt-on-the-ubiquiti-edgerouter-x.html

Also LEDE forum suggests that it can be installed just by using sysupgrade tar
https://forum.lede-project.org/t/ubiquiti-edgerouter-x-factory-bin-image-missing/1458/11

robimarko:
Also rebased this and PR #45 to development

kostko:
> No,it is not build specially,in menu makeconfig you just select ramdisk to be built also and then kernel image as well as factory tar which is not used is generated.

So yes, this is exactly the same as for Mikrotik :-)

robimarko:
Ok,did not know that.
I will check out the code for Mikrotik now.

I see that you use a custom profile.
That profile is being defined where?

kostko:
No, the problem is that for Mikrotik, we had some custom patches so that all images were generated immediately (this was for OpenWrt). So this is something that needs proper handling in the firmware builders before it can be configured in the descriptor. So I think this PR is ok for now.

robimarko:
Ok.
I will look into installation by just using sysupgrade image in following days to see if kernel is even needed as it was for OpenWRT

I will rebase this PR on development as PR #45 was merged

robimarko:
Rebased on development

PR #49, opened 2017-06-01
Update OpenWRT default builders and add LEDE ar71xx builder

robimarko:
Update OpenWRT builders included by default to docker-compose.yml and add LEDE ar71xx builder as one of defaults.


kostko:
Looks good, merged, thanks!

PR #52, opened 2017-06-11
devices: Add Unifi AC Mesh device descriptor

robimarko:
Add device descriptor for Unifi AC Mesh.
It is direct hardware copy from Unifi AP AC Lite and uses same image.

### DIFF
--- a/nodewatcher/modules/devices/ubnt/unifi.py
+++ b/nodewatcher/modules/devices/ubnt/unifi.py
@@ -84,6 +84,15 @@ class UBNTUniFiAPACLite(cgm_devices.DeviceBase):
     }
 
 
+class UBNTUniFiACMesh(UBNTUniFiAPACLite):
+    """
+    UBNT UniFi AC Mesh device descriptor.
+    """
+
+    identifier = 'ub-umesh-ac'
+    name = "UniFi AC Mesh"
+
+
 class UBNTUniFiAPACPro(cgm_devices.DeviceBase):
     """
     Ubiquiti UniFi AP AC Pro device descriptor.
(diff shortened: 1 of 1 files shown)


nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -84,6 +84,15 @@ class UBNTUniFiAPACLite(cgm_devices.DeviceBase):
     }
 
 
+class UBNTUniFiACMesh(UBNTUniFiAPACLite):
+    """
+    UBNT UniFi AC Mesh device descriptor.
+    """
+
+    identifier = 'ub-ac-mesh'
```
kostko: I don't think this identifier is the best. For UniFi AP AC Lite we have `ub-uap-ac-lite`. So for mesh we would want something like `ub-umesh-ac` as otherwise the UniFi part gets lost. What do you think?

robimarko: Sure,that looks good to me.
I have changed it to ub-umesh-ac and squashed.

robimarko:
I have changed identifier to ub-umesh-ac and squashed.

kostko (approved):

nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -84,6 +84,15 @@ class UBNTUniFiAPACLite(cgm_devices.DeviceBase):
     }
 
 
+class UBNTUniFiACMesh(UBNTUniFiAPACLite):
+    """
+    UBNT UniFi AC Mesh device descriptor.
+    """
+
+    identifier = 'ub-umesh-ac'
```
mitar, line 92: I think it should be `ub-uap-ac-mesh`. Because it is still part of `UniFi AP AC` family: https://store.ubnt.com/products/unifi-ac-mesh-ap

nodewatcher/modules/devices/ubnt/unifi.py
```diff
@@ -84,6 +84,15 @@ class UBNTUniFiAPACLite(cgm_devices.DeviceBase):
     }
 
 
+class UBNTUniFiACMesh(UBNTUniFiAPACLite):
```
mitar, line 87: And this `UBNTUniFiAPACMesh`.

robimarko:
Well it is merged now.
If you think that those are better you can change them

PR #54, opened 2017-06-11
devices: Add TL-WR841N v12 device descriptor

robimarko:
Adding device descriptor for TL-WRT841N v12 since it is now only version being sold.
It is identical to v11 and uses the same image but with the different identifier.
Support for it was added in LEDE v17.01.2


kostko (approved):

kostko:
Please rebase.

robimarko:
Rebased onto development

PR #56, opened 2017-06-13
Refactor network defaults

kostko:
Some of the individual configuration classes still need to be moved to their respective modules, but the concept should allow for arbitrary complex defaults, while still being modular.

TODO
* [x] Move individual configuration classes to their respective modules (where such a module is obvious). Some suggestions are in code comments.
* [ ] Think whether a dictionary is the best way of passing context between modules and possibly refactor key names for individual defaults modules.
* [x] Add missing documentation for some defaults modules (define what they export into the context dictionary).
* [ ]  ~~Think about how to add tests for defaults in general.~~


mitar (approved):
This is a bit over my head, but it seems OK.

PR #57, opened 2017-06-25
802.11s integration

robimarko:
Work in progress for adding 802.11s mesh mode.

### DIFF
--- a/nodewatcher/core/generator/cgm/models.py
+++ b/nodewatcher/core/generator/cgm/models.py
@@ -397,7 +397,8 @@ class WifiInterfaceConfig(InterfaceConfig, RoutableInterface, UplinkableInterfac
 
         return list(vifs).index(self.pk)
 
-registration.point('node.config').register_choice('core.interfaces#wifi_mode', registration.Choice('mesh', _("Mesh")))
+registration.point('node.config').register_choice('core.interfaces#wifi_mode', registration.Choice('mesh', _("Mesh (ad-hoc)")))
+registration.point('node.config').register_choice('core.interfaces#wifi_mode', registration.Choice('mesh11s', _("Mesh (802.11s)")))
 registration.point('node.config').register_choice('core.interfaces#wifi_mode', registration.Choice('ap', _("AP")))
 registration.point('node.config').register_choice('core.interfaces#wifi_mode', registration.Choice('sta', _("STA")))
 registration.point('node.config').register_choice('core.interfaces#wifi_bitrates_preset', registration.Choice(None, _("Allow all supported bitrates")))
--- a/nodewatcher/modules/defaults/network_profile/defaults/wireless.py
+++ b/nodewatcher/modules/defaults/network_profile/defaults/wireless.py
@@ -157,12 +157,14 @@ class WirelessModule(NetworkModuleMixin, registry_forms.FormDefaultsModule):
                     cgm_models.WifiInterfaceConfig,
                     parent=wifi_radio,
                     configuration={
-                        'mode': 'mesh',
+                        # Use 802.11s if checked, otherwise use ad-hoc.
+                        'mode': 'mesh11s' if 'wifi-mesh-80211s' in network_profiles else 'mesh'
                         'essid': get_project_ssid('mesh'),
                         'bssid': get_project_ssid('mesh', attribute='bssid'),
                         'routing_protocols': routing_protocols,
                     },
                 )
+
         elif node_type == 'backbone':
             # If the device is of type "Backbone", create one AP/STA.
             if wifi_backbone_defaults is not None:
--- a/nodewatcher/modules/defaults/network_profile/models.py
+++ b/nodewatcher/modules/defaults/network_profile/models.py
@@ -36,5 +36,6 @@ registration.point('node.config').register_choice('network.profile#profiles', re
 registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('wifi-uplink', _("Use a wireless interface for uplink")))
 registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('no-wifi-ap', _("Disable wireless AP")))
 registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('no-wifi-mesh', _("Disable wireless mesh")))
+registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('wifi-mesh-80211s', _("Use 802.11s for meshing")))
 registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('hostname-essid', _("Use hostname as ESSID")))
 registration.point('node.config').register_item(NetworkProfileConfig)
--- a/nodewatcher/modules/platforms/openwrt/cgm/network.py
+++ b/nodewatcher/modules/platforms/openwrt/cgm/network.py
@@ -789,6 +789,12 @@ def network(node, cfg):
                     # Override default mcast_rate to avoid broadcast traffic from
                     # stealing too much air time
                     wif.mcast_rate = 6000
+                elif vif.mode == 'mesh11s':
+                    wif.mode = 'mesh'
+                    # Use BSSID from Mesh as Mesh ID.
+                    wif.mesh_id = vif.bssid
+                    # Use full wpad instead of wpad-mini because of 802.11s Mesh.
+                    cfg.packages.update(['-wpad-mini', 'wpad'])
                 elif vif.mode == 'sta':
                     wif.mode = 'sta'
 


nodewatcher/core/generator/cgm/models.py
```diff
@@ -398,6 +398,7 @@ def get_index(self):
         return list(vifs).index(self.pk)
 
 registration.point('node.config').register_choice('core.interfaces#wifi_mode', registration.Choice('mesh', _("Mesh")))
+registration.point('node.config').register_choice('core.interfaces#wifi_mode', registration.Choice('802.11s', _("802.11s Mesh")))
```
kostko: Perhaps we should rename both mesh modes as follows:
* For existing ad-hoc, we can change the description to `"Mesh (ad-hoc)"` and leave the `mesh` identifier.
* For the new 802.11s, give it identifier `mesh11s` and change the description to `"Mesh (802.11s)"`.

nodewatcher/modules/defaults/network_profile/defaults/wireless.py
```diff
@@ -163,6 +163,21 @@ def get_project_ssid(purpose, attribute='essid'):
                         'routing_protocols': routing_protocols,
                     },
                 )
+
+            # Create 802.11s mesh interface.
+            if 'no-wifi-mesh' and '802.11s-wifi-mesh' in network_profiles:
```
kostko: This check doesn't seem right to me, so `if 'no-wifi-mesh'` will always evaluate to True so it will be ignored (you probably meant `if 'no-wifi-mesh' in network_profiles and ...`).

Also, I think it would be better to move that into the above if, so you have something like this:
```python
if 'no-wifi-mesh' not in network_profiles:
    if '802.11s-wifi-mesh' in network_profiles:
        # Configure 802.11s.
    else:
        # Configure ad-hoc mesh.
```

nodewatcher/modules/defaults/network_profile/defaults/wireless.py
```diff
@@ -163,6 +163,21 @@ def get_project_ssid(purpose, attribute='essid'):
...
+                self.setup_interface(
+                    state,
+                    cgm_models.WifiInterfaceConfig,
+                    parent=wifi_radio,
+                    configuration={
+                        'mode': '802.11s',
+                        'essid': get_project_ssid('802.11s'),
```
kostko: I don't think we need a special ESSID for 802.11s? I think this should be unified with normal ad-hoc mesh or will this cause any problems?

nodewatcher/modules/defaults/network_profile/defaults/wireless.py
```diff
@@ -163,6 +163,21 @@ def get_project_ssid(purpose, attribute='essid'):
...
+                        'mesh_key': get_project_ssid('802.11s_mesh_id'),
```
kostko: This is probably not ok as the model doesn't have a `mesh_key` field. Why don't you just set `bssid` and then use `bssid` as `mesh_id` when configuring UCI in CGMs?

nodewatcher/modules/defaults/network_profile/models.py
```diff
@@ -36,5 +36,6 @@ class RegistryMeta:
 registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('wifi-uplink', _("Use a wireless interface for uplink")))
 registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('no-wifi-ap', _("Disable wireless AP")))
 registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('no-wifi-mesh', _("Disable wireless mesh")))
+registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('802.11s-wifi-mesh', _("Enable 802.11s wireless mesh")))
```
kostko: I think a better name would be `wifi-mesh-80211s`. And the description more like `"Use 802.11s for meshing"`.

nodewatcher/modules/defaults/network_profile/defaults/wireless.py
kostko: Oh, I see that below you already use `bssid`. So you should just change this.

nodewatcher/modules/platforms/openwrt/cgm/network.py
```diff
@@ -789,6 +789,14 @@ def network(node, cfg):
                     # Override default mcast_rate to avoid broadcast traffic from
                     # stealing too much air time
                     wif.mcast_rate = 6000
+                elif vif.mode == '802.11s':
+                    wif.mode = 'mesh'
+                    # Use BSSID from Mesh as Mesh ID
+                    wif.mesh_id = vif.bssid
+                    # Use full wpad instead of wpad-mini because of 802.11s Mesh.
+                    cfg.packages.update([
+                    'wpad -wpad-mini'
```
kostko: Please don't specify multiple packages in the same string. Use two separate entries for that, for example:
```python
cfg.packages.update(['-wpad-mini', 'wpad'])
```

nodewatcher/modules/platforms/openwrt/cgm/network.py
```diff
@@ -789,6 +789,14 @@ def network(node, cfg):
                     # Override default mcast_rate to avoid broadcast traffic from
                     # stealing too much air time
                     wif.mcast_rate = 6000
+                elif vif.mode == '802.11s':
+                    wif.mode = 'mesh'
+                    # Use BSSID from Mesh as Mesh ID
```
kostko: Dot at the end of comment.

kostko (changes requested):
Nice work, I added some comments.

robimarko:
@kostko I have made requested changes.
Now the issue is that it does not work yet.
Also When 802.11s is ticked then if the device has USB port specified it automatically creates wireless radio from USB port and no 802.11s wireless interface

kostko:
> if the device has USB port specified it automatically creates wireless radio from USB port

This has nothing to do with this PR and is caused by the new automatic configuration of multiple radios. I will fix it so it will never automatically configure USB radios.

nodewatcher/modules/defaults/network_profile/defaults/wireless.py
```diff
@@ -152,17 +152,33 @@ def get_project_ssid(purpose, attribute='essid'):
...
-                    configuration={
-                        'mode': 'mesh',
-                        'essid': get_project_ssid('mesh'),
-                        'bssid': get_project_ssid('mesh', attribute='bssid'),
-                        'routing_protocols': routing_protocols,
-                    },
-                )
+                if 'wifi-mesh-80211s' in network_profiles:
+                    # Configure 802.11s mesh interface.
+                        self.setup_interface(
```
kostko: This indentation doesn't seem ok. BTW, I see that now both branches are almost the same, only the `mode` is different. Perhaps we could just unify both and have something like this instead:
```python
configuration={
    'mode': 'mesh11s' if 'wifi-mesh-80211s' in network_profiles else 'mesh',
    # ...
```

robimarko: Well, that would not simplify thing a lot.
What is wrong with identation?

kostko: There seem to be an extra 4 spaces?

kostko:
I've pushed some USB and other fixes to development.

nodewatcher/modules/defaults/network_profile/defaults/wireless.py
robimarko: Hm, I did not see that.
The previous indentation that I made was giving me PEP8 errors.
This one does not, but the code does not work.

robimarko:
Ok,I will then rebase this on development

robimarko:
Now mesh interface is configured.

nodewatcher/core/generator/cgm/models.py
```diff
@@ -397,7 +397,8 @@ def get_index(self):
 
         return list(vifs).index(self.pk)
 
-registration.point('node.config').register_choice('core.interfaces#wifi_mode', registration.Choice('mesh', _("Mesh")))
+registration.point('node.config').register_choice('core.interfaces#wifi_mode', registration.Choice('mesh', _("Mesh (ad-hoc)")))
```
mitar, line 400: Can we rename this to `mesh-ah` and the other one to `mesh-11s` or something? But then we need a migration. So maybe not. :-)

nodewatcher/modules/defaults/network_profile/models.py
```diff
@@ -36,5 +36,6 @@ class RegistryMeta:
 registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('wifi-uplink', _("Use a wireless interface for uplink")))
 registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('no-wifi-ap', _("Disable wireless AP")))
 registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('no-wifi-mesh', _("Disable wireless mesh")))
+registration.point('node.config').register_choice('network.profile#profiles', registration.Choice('wifi-mesh-80211s', _("Use 802.11s for meshing")))
```
mitar, line 39: @kostko, could we have here maybe `wifi-mesh-80211s` and `wifi-mesh-adhoc`? Is it possible to have profiles enabled by default? So that we would have `wifi-mesh-adhoc` by default enabled, for legacy?

If not, I would label this option as `Use 802.11s for meshing instead of ad-hoc`.

mitar:
This really looks good. I think it is important to be able to experiment with other things. But yes, let's add it as an alternative to ad-hoc and not replace it.

nodewatcher/modules/defaults/network_profile/models.py
robimarko: Well, ad-hoc is still enabled by default unless you tick Disable wireless mesh or Use 802.11s.
So there is no need to change anything for now

mitar: And I am asking if we can have something ticked by default. Because then it would be easier to understand these options if you would have "enable mesh with ad-hoc" and "enable mesh with 802.11s".

mitar:
BTW, can both ad-hoc and 802.11s be enabled at the same time?

nodewatcher/modules/defaults/network_profile/models.py
kostko: It is possible to enable by default but for existing nodes we would need a migration. A bigger problem is that then you would be able to check both options.

The best solution would be to instead introduce an additional field in the network profile model called `mesh_mode` or something, which would instead be a single choice field. And then use that in defaults.

kostko:
> BTW, can both ad-hoc and 802.11s be enabled at the same time?

I am not sure if we want that.

robimarko:
Technically yes, but that would cripple performance for sure.

nodewatcher/modules/defaults/network_profile/models.py
robimarko: I think that it is not bad that ad-hoc is still active but default and you can disable it or change to 802.11s by a simple tick.
No need to complicate things.

mitar:
> I am not sure if we want that.

Does it not work? Or you think it will just be useless use of spectrum? I see it that if there are no neighbors, then it does not really consume anything. But if there are ad-hoc neighbors, then it would still work. Maybe we should just make sure that if two nodes see each other on both ad-hoc or 802.11s, user can choose which one to prefer.

nodewatcher/core/generator/cgm/models.py
robimarko: It can be refactored but I think it would require DB migration.

mitar:
> Technically yes, but that would cripple performance for sure.

Why exactly? Please make better arguments. If interface is up and is not used, I do not see why it would impact performance a lot? Maybe just for beacons sent? That should not be too much.

nodewatcher/modules/defaults/network_profile/models.py
mitar: Yes. We could have one field, mesh: and you can select: ad-hoc, 802.11s, or none.

nodewatcher/core/generator/cgm/models.py
mitar: DB migrations are OK to do. :-)

@kostko, what do you think?

robimarko:
Well even if the interface is not being used it is still up will impact performance.
But in order to determine how much field test should be conducted.

nodewatcher/modules/defaults/network_profile/models.py
robimarko: Well,that would be great

valentt:
@mitar I understand your point for having both (if possible) mesh interfaces because that would keep compatibility and make migration smoother. We had similar migration with routing protocols and kept both in parallel for some time. So having both adhoc modes make sense from that point, to prevent people from breaking their network by updating one node that then others can't connect .

But because of that I would suggest to keep adhoc as default mesh protocol and have option to switch to 802.11s if somebody wants to test and deploy 802.11s mesh, but not to have both protocols at the same time. It just complicates things too much and will confuse people as well (two mesh ssids).

valentt:
And then we can test if needed how it works if we have two mesh interfaces.

mitar:
I think this is why I think current proposal is to have one choice to select for mesh: no mesh, ad-hoc mesh, 802.11s mesh, if we want to support only one.

But the argument for having both is not for legacy purposes. I will discuss this more in chat. But in short, ad-hoc mesh serves a different purpose than 802.11s mesh. 802.11s mesh is a L2 mesh, it makes maybe easier to cover a local area you want to cover with APs, but you do not have cables to every location. ad-hoc mesh is meant to create a country wide network. A network which works (but not necessary efficiently) when the normal Internet fails. 802.11s does not scale to more than 32 nodes, it not even in the standard to support bigger networks, to what I remember.

So the proper approach here would be probably to have both of those enabled, so that if Internet uplink fails, connectivity is established by ad-hoc mesh. It could also be automatic process: if Internet uplink fails (which means also over 802.11s there is no Internet), turn off 802.11s and switch to ad-hoc. The reason is because maybe there people will start bringing more ad-hoc devices to heal bad spots in the network and with 802.11s the mesh would melt.

Anyway, I think that for experimentation it is good to support also 802.11s. But we should not remove ad-hoc mesh just because we are experimenting with 802.11s. One important part of wlan slovenija network is that under the table we are building an autonomous independent network. That you can just throw devices out and they just connect. Badly, not efficiently, but connect.

kostko:
> But we should not remove ad-hoc mesh just because we are experimenting with 802.11s.

Where did you see that we are removing ad-hoc? This PR just adds 802.11s as an optional enable.

mitar:
> Where did you see that we are removing ad-hoc? This PR just adds 802.11s as an optional enable.

I am talking about removing ad-hoc from nodes, not from support. I am talking about the way Valent is proposing to configure their nodes.

Anyway, let's experiment and let's see.

robimarko:
Well, ad-hoc would be removed only if the user who is configuring node wants to do so.
He can choose to have ad-hoc or 802.11s as Mesh solution or chose to not have Mesh at all.

I think that that is logical way

nodewatcher/modules/defaults/network_profile/models.py
kostko: Yes, this is easily added with suitable migrations once 802.11s is finalized and confirmed to be working correctly.

nodewatcher/core/generator/cgm/models.py
kostko: Yes, we can add migrations and refactor this, but let's test 802.11s first.

robimarko:
I generated a new image and flashed it to see if mesh_id will be configured correctly.
Unfortunately, it is completely empty in configuration.
I can't seem to find the issue.

config wifi-iface
        option device 'radio0'
        option encryption 'none'
        option ssid 'mesh.otvorenamreza.org'
        option mode 'mesh'
        option mesh_id ''
        option ifname 'mesh11s0'
        option network 'mesh11s0'

kostko:
> Unfortunately, it is completely empty in configuration.

It is set, but to an empty string. So this means that `wif.mesh_id = vif.bssid` is executed correctly, so it must be that `vif.bssid` is empty. Can you check in your node's configuration if the 802.11s interface has a BSSID set?

robimarko:
I just checked and BSSID is empty.
Which is weird since ad-hoc mesh has BSSID field populated and that same BSSID should be used as set in code

robimarko:
I have solved empty mesh_id.
Field name in wireless configuration was wrong.

Now everything is configured properly.

kostko:
That's great! Can you unify the part of the code that configures the mesh interface as previously discussed in the comments?

robimarko:
Sure, after it is tested to work.
It should be tested today and if it works I will unify it and remove docker-compose.yml and settings.py changes and then squash this to be mergeable

kostko:
Sounds good! :+1:

robimarko:
Just tested and it works.
I did not test speeds but we will test those out too.

I will unify the code today and then squash to be ready for merging

robimarko:
I have unified the code.
I will squash it and have it ready for merging.

robimarko:
Code is squashed and ready.

nodewatcher/core/generator/cgm/models.py
```diff
@@ -397,7 +397,8 @@ def get_index(self):
...
+registration.point('node.config').register_choice('core.interfaces#wifi_mode', registration.Choice('mesh11s', _("Mesh (802.11s)")))
```
kostko: Can you change this to `mesh-11s`?

nodewatcher/modules/defaults/network_profile/defaults/wireless.py
```diff
@@ -157,12 +157,14 @@ def get_project_ssid(purpose, attribute='essid'):
                     cgm_models.WifiInterfaceConfig,
                     parent=wifi_radio,
                     configuration={
-                        'mode': 'mesh',
+                        # Use 802.11s if checked, otherwise use ad-hoc
```
kostko: Add dot at the end.

kostko (approved):
One minor comment, otherwise looks good to merge.

nodewatcher/modules/defaults/network_profile/defaults/wireless.py
robimarko: Done, squashed and ready.

PR #60, opened 2017-07-01
WIP: Integrate SNMP Daemon configuration

robimarko:
Work in progress for adding SNMP Daemon configuration to Nodewatcher.
That way we would be able to configure snmpd as an extra package instead of having to manually install and configure it later.

Currently, I think that most of the work is done, but database migrations need to be added as well some syntax fixes.

Note: Ignore parts about settings.py and docker-compose as well beta builders.
Those builders are the same as current ones but with snmpd package.
They are available here: https://hub.docker.com/r/otvorenamreza/lede-builder/tags/


nodewatcher/modules/monitor/sources/snmp/apps.py
```diff
@@ -0,0 +1,7 @@
+from django import apps
+
+
+class SNMPDConfig(apps.AppConfig):
```
kostko: I think this should be named `SNMPConfig` as the module (and the protocol) is SNMP. Also, fix above in `__init__.py`.

nodewatcher/modules/monitor/sources/snmp/apps.py
```diff
@@ -0,0 +1,7 @@
...
+    name = 'nodewatcher.modules.monitor.sources.snmp'
+    label = 'snmpd'
```
kostko: Same as above.

nodewatcher/modules/monitor/sources/snmp/cgm.py
```diff
@@ -0,0 +1,39 @@
...
+    Configures SNMP daemon.
+    """
+
+    try:
+        pkgcfg = pkgcfg.get()
+    except cgm_models.PackageConfig.MultipleObjectsReturned:
+        raise cgm_base.ValidationError(_("Only one SNMP daemon instance can be configured."))
+
+    # Configure SNMP daemon.
+    sysname = node.config.core.general().get_name()
```
kostko: Perhaps just use `name` instead of `sysname`?

nodewatcher/modules/monitor/sources/snmp/models.py
```diff
@@ -0,0 +1,27 @@
...
+from nodewatcher.core.generator.cgm import models as cgm_models
+
+
+class SNMPDaemonConfig(cgm_models.PackageConfig):
+    """
+    SNMP Daemon configuration.
+    """
+
+    community = models.CharField(max_length=100)
+    sysLocation = models.CharField(max_length=100)
```
kostko: Please rename this to `sys_location` (e.g., no camel case in variable names). Also, any specific reason why this is not just `location`? Also, can this be inferred from somewhere else (for example we actually have the node's location already available in the location module, so perhaps that information can be reused and this field is not needed at all)?

nodewatcher/modules/monitor/sources/snmp/models.py
```diff
@@ -0,0 +1,27 @@
...
+    sysContact = models.CharField(max_length=100)
```
kostko: Please rename this to `sys_contact` (e.g., no camel case in variable names). Also, any specific reason why this is not just `contact`? Also, can this be inferred from somewhere else (for example we actually have the node's contact already available as the maintainer, so perhaps that information can be reused and this field is not needed at all)?

nodewatcher/settings.py
```diff
@@ -354,6 +354,9 @@ def user_url(user):
 
     # Accounts support.
     'nodewatcher.extra.accounts',
+
+    # SNMP Daemon.
+    'nodewatcher.modules.monitor.sources.snmp',
```
kostko: I think this should be moved above (before `nodewatcher.extra.*` start).

nodewatcher/settings.py
```diff
@@ -726,4 +729,4 @@ def user_url(user):
 
 # Allowed hosts (required for production use, when DEBUG is false). Set it
 # to the (virtual HTTP) hostname under which you have nodewatcher installation.
-ALLOWED_HOSTS = []
+ALLOWED_HOSTS = ['212.237.52.226']
```
kostko, line 735: Please remove this as this is something related to your development environment.

kostko (changes requested):

kostko:
Nice work! I've added some comments. I think that we should reuse as much information as possible from other modules (e.g., location and node maintainer), so you don't need to enter the same information multiple times.

kostko:
> Currently, I think that most of the work is done, but database migrations need to be added as well some syntax fixes.

Yes, please commit database migrations as well.

nodewatcher/modules/monitor/sources/snmp/cgm.py
```diff
@@ -0,0 +1,39 @@
...
+    system.sysName = sysname
+    system.sysLocation = pkgcfg.sysLocation
+    system.sysContact = pkgcfg.sysContact
+    system.sysObjectID = node.uuid
+
+    public = cfg.snmpd.add(com2sec='public')
+    public.community = pkgcfg.community
+
+    exec = cfg.snmpd..add(**{'exec': 'exec'})
+    exec['name'] = 'filedescriptors'
```
kostko: This is not needed, you can do `exec.name = ...`, also below.

nodewatcher/settings.py
robimarko: Yeah,that will be removed as soon this works.

nodewatcher/modules/monitor/sources/snmp/cgm.py
robimarko: Ok,thanks.
Will change that.

robimarko:
Requested changes are made.
I will now prepare database migrations.

nodewatcher/modules/monitor/sources/snmp/cgm.py
robimarko: Well to me it really does not matter so I changed it to name

nodewatcher/settings.py
robimarko: I agree,change made

nodewatcher/modules/monitor/sources/snmp/apps.py
robimarko: I agree,change made

nodewatcher/modules/monitor/sources/snmp/apps.py
robimarko: I agree,change made

nodewatcher/modules/monitor/sources/snmp/models.py
robimarko: I agree,changed to location.

nodewatcher/modules/monitor/sources/snmp/models.py
robimarko: I agree,changed to contact.

docker-compose.yml
```diff
@@ -12,11 +12,11 @@ services:
             PGSQL_DB_1_POSTGIS: "true"
         volumes:
             # Use a volume to enable database persistence across restarts
-            - /tmp/nodewatcher-db:/var/lib/postgresql/9.5/main
+            - ~/nodewatcher-db:/var/lib/postgresql/9.5/main
```
mitar: Why this change? Because it is reset between reboots of host laptop? But I do not like that now developers would have a random directory in home directory. One option could be to make this [configurable through variable substitution](https://docs.docker.com/compose/compose-file/#variable-substitution). And leave `/tmp` as default.

mitar, line 15: Why this change? Because it is reset between reboots of host laptop? But I do not like that now developers would have a random directory in home directory. One option could be to make this [configurable through variable substitution](https://docs.docker.com/compose/compose-file/#variable-substitution). And leave `/tmp` as default.

mitar: I know that you asked to ignore this change, but let's maybe fix this for you if this is a need you have?

nodewatcher/settings.py
```diff
@@ -354,6 +357,7 @@ def user_url(user):
 
     # Accounts support.
     'nodewatcher.extra.accounts',
+
```
mitar: Unnecessary change.

mitar (changes requested):
As mentioned in the chat, I think that we should try to fetch information for community, location, and contact from the rest of nodewatcher.

If the rest of nodewatcher does not have enough values you need, then add it to nodewatcher and then allow selection and reasonable defaults.

For community, you could pull data from:
* nodewatcher installation name (if snmp is used per whole nodewatcher installation)
* project
* django group name (not yet very exposed in nodewatcher, but already available)
* location area (a new concept which would be between projects which are not geographically limited, and location address, which is limited to one address)

And location, you could use:
*  project
* django group name (if it represent a location, like name of a hackerspace)
* location area
* location address

nodewatcher/settings.py
robimarko: Well Kostko asked for it.

docker-compose.yml
robimarko: Because in order to test this out properly I cant have my DB reset after every boot.
I will revert that back,this way it is easier for me to test out since all the code is equal on remote and local branch

robimarko:
Those are good ideas but over complicate this currently.
After those are added to Nodewatcher changes can be made to apply them to SNMP module as well.
We wouldn't use project name as the community as we use our own community name for all of the nodes if it were to change I bet you that soon we would not know which node is in which community.

I have already explained why location does not suit us as well in chat

nodewatcher/settings.py
mitar: OK. :-)

docker-compose.yml
mitar: OK, but read my proposal for the better solution as well and consider making a change to fix this so that in the future you do not have to commit in temporary changes for your development to work.

Also, why do you have to commit this in? You can have local changes to files which you do not commit. git allows partial commits. Maybe use some GUI like GitHub client which allows an easy way to select what you want to commit.

nodewatcher/modules/monitor/sources/snmp/cgm.py
```diff
@@ -0,0 +1,39 @@
...
+    system = cfg.snmpd.add('system')
+    system.sysName = name
+    system.sysLocation = pkgcfg.location
+    system.sysContact = pkgcfg.contact
+    system.sysObjectID = node.uuid
+
+    public = cfg.snmpd.add(com2sec='public')
+    public.community = pkgcfg.community
+
+    execute = cfg.snmpd.add(**{'execute': 'exec'})
```
kostko: I don't think this is correct anymore as you also changed the UCI type. You should leave the dictionary as `**{'exec': 'exec'}`, it is just the variable that has to be `execute` instead of `exec`.

nodewatcher/modules/monitor/sources/snmp/cgm.py
```diff
@@ -0,0 +1,39 @@
...
+    execute.name = 'filedescriptors'
+    execute.prog = '/bin/cat'
+    execute.args = '/proc/sys/fs/file-nr'
+
+    # Ensure that "snmpd" package is installed
```
kostko: Add dot at the end.

nodewatcher/modules/monitor/sources/snmp/cgm.py
kostko: Also, based on the stuff below, why is the section named `exec`? Wouldn't a better name be something like `filedescriptors`, for example can you test if this works:
```
**{'exec': 'filedescriptors'}
```

nodewatcher/modules/monitor/sources/snmp/cgm.py
```diff
@@ -0,0 +1,39 @@
...
+    system.sysContact = pkgcfg.contact
+    system.sysObjectID = node.uuid
+
+    public = cfg.snmpd.add(com2sec='public')
+    public.community = pkgcfg.community
+
+    execute = cfg.snmpd.add(**{'execute': 'exec'})
+    execute.name = 'filedescriptors'
+    execute.prog = '/bin/cat'
+    execute.args = '/proc/sys/fs/file-nr'
```
kostko: BTW, why is this "number of file descriptors" endpoint hardcoded here? Is this something that you are using? In this case it should probably be configurable through nodewatcher?

nodewatcher/modules/monitor/sources/snmp/models.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class SNMPDaemonConfig(cgm_models.PackageConfig):
+    """
+    SNMP Daemon configuration.
+    """
+
+    community = models.CharField(max_length=100, verbose_name=_("SNMP Community"))
+    location = models.CharField(max_length=100, verbose_name=_("Device location"))
+    contact = models.CharField(max_length=100, verbose_name=_("Contact info"))
```
kostko: Change verbose name to just `"Contact"`.

nodewatcher/settings.py
kostko: Not for the blank line, I just asked you to move what was previously there? Please remove the blank line.

kostko:
BTW, I just now noticed that you have this under `monitor.sources`, but actually you only provide reporting and not a SNMP source for nodewatcher. Are you planning to also implement SNMP as a source in nodewatcher? This would of course only make sense for including devices, which don't support `nodewatcher-agent`?

robimarko:
@mitar Suggested that it should be placed here.
This way in future it can be extended to be the source for Nodewatcher.
I currently don't have plans for that.

nodewatcher/modules/monitor/sources/snmp/models.py
robimarko: Ok,done.

nodewatcher/settings.py
robimarko: You are right,blank line removed

nodewatcher/modules/monitor/sources/snmp/cgm.py
robimarko: It would, but I just exec since that is what it is named in default SNMPD configuration.
But like you said filedescriptors is more appropriate.

nodewatcher/modules/monitor/sources/snmp/cgm.py
robimarko: Because this is used by default and returns information about the whole filesystem.
We use this as this is really useful

docker-compose.yml
robimarko: This is not an issue for me.

I will look into partial commiting

nodewatcher/modules/monitor/sources/snmp/cgm.py
robimarko: Added

docker-compose.yml
akrpic77: This could be ./nodewatcher-db so it's relative to docker-compose.yml

robimarko: Well,that is actually really good idea.

mitar:
> Those are good ideas but over complicate this currently.

Why it complicate? It makes it easier to manage and change multiple nodes at once. If you will ever have to change anything you will be able to change it in one place.

> We wouldn't use project name as the community as we use our own community name for all of the nodes if it were to change I bet you that soon we would not know which node is in which community.

Yes, different users of nodewatcher would have different needs. Some could configure it so that all nodes of nodewatcher is in one community, some could configure it so that each project is one community.

Now, if you are saying that there is no point in having multiple SNMP communities in the network, then I propose we do not allow maintainers to configure random community values here and this is something directly configured by nodewatcher for the whole network managed by nodewatcher for all nodes. In the case of *wlan slovenija* installation of nodewatcher, that would then be `wlan-si.net` or some other related name of the community. But because I assume you would want a name for your nodes to match your project, I would propose that we also allow projects to have their own SNMP communities.

> I have already explained why location does not suit us as well in chat

Yes. I am not saying that you should use location, but that location could be one of options to use for users who do want it. For you a wider concept like "location area" might be a better one to choose.

You have to remember that nodewatcher is not a centralized web interface to UCI. So the idea is not to have simply what would run on one node through web interface to set those values, but on a central server. The idea is that you have a system to manage nodes. And managing means that you do not have to go around 100 nodes when you change one value, but you can do that in one place, and it updates values everywhere. This is why it is good to have values being defined only in one place, so that they can be reused around. Look at form defaults. Yes, you can disable defaults and enter everything manually, but the question is what happens when you want to update some value.

@kostko, what do you think, is this maybe something better resolved at the level of form defaults? So a community can define their own defaults for SNMP, and then they can in defaults say from where those values come from. In form defaults, if you change code which generates where form defaults are getting values, does this automatically update all configurations of nodes, or just the next time you are updating configuration using defaults?

BTW, what I am thinking here would be that it would be great if SNMP would work the same as nodewatcher agent. So that it could be a replacement for it. So that it is an integral part of nodewatcher, which can just work out of the box. So my question is, how much configuration do we even want to expose to the user. Or should be only to choose which protocol to use for getting data (nodewatcher pull, push, or snmp). And then the rest would simply be chosen by nodewatcher?

@robimarko: Why you want SNMP support? So that nodewatcher would be collecting information from nodes using SNMP? Or that some other programs to monitor could also monitor nodes? And then you want to expose information from nodewatcher like community and location to those systems? I think it the latter case it is then even more important to dynamically populate those values, so that it is also a form of syncing/federation between nodewatcher and that other system.

docker-compose.yml
mitar: And we could also add `nodewatcher-db` to `.gitignore`.

nodewatcher/modules/monitor/sources/snmp/cgm.py
mitar: I opened also: https://dev.wlan-si.net/ticket/1438

docker-compose.yml
akrpic77: ... and to .dockerignore to reduce build times.

robimarko:
Mitar I can mostly agree.
But we differ in some views.
Forcing people to use project name for SNMP community is not preferred.
That way use case for SNMP is being limited.
I think that it would be good to have project name being default SNMP community and that users can change it in node settings if it does not suit their needs.
A better solution would be having the option to set per project community name that does not have to be equal to the project name.

For location, I think that good solution would be having the option to make groups withing Projects and that than group name or group location would be default location with having option to change it in node settings.

I agree that this would be better solved by having per project or per group default settings that still can be changed in node settings.

@mitar We want SNMP support because we use it every day for monitoring 92 nodes currently.
We currently use LibreNMS for that purpose.
It offers us email, rocket chat and various other alerts that are fully customizable.
Also, offers integration with various applications and has a lot of extensions.
And it enables us to monitor parts of our backbone that do not have Nodewatcher firmware because either LEDE/OpenWRT do not support them or performance penalty is too high.
Mostly that is a lot of PtP links using Ubiquiti AC gear and plenty of POE switches as others like Mikrotik and everything else not running Nodewatcher firmware.
And numerous other features.

It would be great that Nodewatcher could use SNMP as data source and that certainly can be something on TODO list for future because SNMP is a lot more difficult to implement than HTTP.
Implementation in Nodewatcher can be done via using LibreNMS API or by simply reading measurements from InfluxDB since LibreNMS natively can export data to InfluxDB and we currently do that for our Grafana graphing.
But having SNMP daemon properly configured is the first step to it.

I look forward to more discussion and further improvement of Nodewatcher as it looks like it is getting better and better every day

kostko:
> Forcing people to use project name for SNMP community is not preferred.

I don't think Mitar was suggesting that, he said that it would be configurable.

> Implementation in Nodewatcher can be done via using LibreNMS API or by simply reading measurements from InfluxDB since LibreNMS natively can export data to InfluxDB and we currently do that for our Grafana graphing.

No, this would be a bad implementation. We don't need additional dependencies (e.g., we will not run LibreNMS just to have SNMP support), so the SNMP source should be stand-alone.

robimarko:
Ok, then we agree on almost all.
I know that it would be bad implementation, but otherwise writing standalone support would be a lot of work.

Something like PySNMP or Easy SNMP can be used but it would still require a lot of work.
Surely something that can be done in future

kostko:
> @kostko, what do you think, is this maybe something better resolved at the level of form defaults? So a community can define their own defaults for SNMP, and then they can in defaults say from where those values come from. In form defaults, if you change code which generates where form defaults are getting values, does this automatically update all configurations of nodes, or just the next time you are updating configuration using defaults?

So defaults are the most flexible (e.g., you can combine multiple sources or make everything static) and this is something that could be handled there. And then we can introduce a defaults module, which configures this based on project or other factors and also allows overrides.

> Something like PySNMP or Easy SNMP can be used but it would still require a lot of work. Surely something that can be done in future

Yes, so I think that adding SNMP source support to nodewatcher would be great so we can use nodewatcher also for devices, which cannot run our agent instead of requiring external systems like LibreNMS to be used.

robimarko:
@kostko I agree completely

mitar:
OK, so what are then next steps? Maybe:
* change this pull request to not be about SNMP source module, but just a SNMPd package with configuration, where you can configure anything you want as community and location
* you can then use defaults to populate community and location with some defaults based on other aspects of nodewatcher
* SNMP source module then configures SNMPd to a common network-wide community, location is not really important in that case, because it would be just to collect information using SNMP from devices in the network
* SNMP source module then adds extends nodewatcher monitor to collect also information from that SNMP community in the network, and adds it to nodewatcher data

robimarko:
I think you pretty much summed it up.
Most difficult part is the last one.

mitar:
I am not sure why would it be too hard? We already have whole infrastructure for running long-lived processes available. You just instead of doing periodic HTTP requests, use PySNMP to fetch SNMP data from nodes. (Or you configure it to listen for SNMP data from nodes.) But yes, we would have to learn how to do it. :-)

robimarko:
Because SNMP despite having simple in its name is nowhere near simple as current HTTP JSON data source.
It is based on tables and on Linux as well as OpenWRT/LEDE by default a lot of stuff that you currently pull with HTTP is not pulled by default but requires the use of exec and custom tables.

If you want to see how data fetched from SNMP looks I can make you account to our LibreNMS monitoring

mitar:
Maybe try to use [this example](http://pysnmp.sourceforge.net/quick-start.html#fetch-snmp-variable) to fetch any value from your existing nodes. You can also [check this project](https://github.com/ninuxorg/netengine/blob/master/netengine/backends/snmp/airos.py) to see various IDs to use to fetch data from AirOS. Once you have those IDs, it is pretty easy to fetch them and store them in database (the latter part is already there).

See [this processor](https://github.com/wlanslovenija/nodewatcher/blob/development/nodewatcher/modules/monitor/sources/http/processors.py#L58). It runs and uses HTTP, but similarly we could use SNMP.

robimarko:
LibreNMS GitHub is an awesome place to look for examples.
They pretty much support everything that has SNMP daemon.

robimarko:
I am just trying out if this works.
Unfortunately,it looks like a problem with writing to DB.
Here is full log:
https://gist.github.com/robimarko/bc6003771467ce5c0213e8ef908aa249

PR #61, opened 2017-07-12
nodeupgrade: Customize to LEDE

robimarko:
Customize template to LEDE instead of using OpenWRT one.
That can cause confusion in some people


kostko (approved):
Looks good, just rebase and we can merge.

robimarko:
Rebased onto development done.

PR #62, opened 2017-08-27
Update USB packages to support USB3.0

robimarko:
I am working on adding a device with USB3.0 to Nodewatcher after next LEDE stable is out, it has USB3.0 port so proper package to support it is needed.
This way if usb3 in device descriptor is also True kmod-usb3 package will be installed instead of kmod-usb2


nodewatcher/modules/platforms/openwrt/cgm/general.py
```diff
@@ -66,7 +66,16 @@ def usb(node, cfg):
         # Include base USB packages for devices supporting USB.
         # TODO: Perhaps this should be made device-specific so only the needed packages are installed.
         cfg.packages.update([
-            'kmod-usb2',
             'kmod-usb-ohci',
             'kmod-usb-uhci',
         ])
+        # Include packages to support USB3.0 if device support it.
+        if device.usb3:
```
kostko, line 73: Where did you get this `device.usb3` from? You need to add support for this to device descriptors, otherwise this will not work.

And I also don't like having two flags (`usb` and `usb3`), it would be much better to have something like this in the device descriptors:
```python
# Device does not support USB (can be either of these).
usb = False
usb = None
# Device supports USB2.
usb = cgm_devices.USB_2
# Device supports USB3.
usb = cgm_devices.USB_3
```

kostko (changes requested):

nodewatcher/modules/platforms/openwrt/cgm/general.py
robimarko: Ok, I like your idea.
Where are those flags defined?

kostko: In `nodewatcher.core.generator.cgm.devices`.

robimarko: Finally looked into cgm.devices, and I don't think I understood your idea.
This way also we need 2 flags

kostko: What do you mean two flags? You only need one `usb` attribute/flag in the device descriptor instead of the recommended two (`usb` and `usb3`). For example I do not want to have something like this in the device descriptor for a device, which supports USB3:
```python
class MyShinyDevice(cgm_devices.DeviceBase):
    """
    My shiny device v1 device descriptor.
    """

    identifier = 'my-shinyv1'
    name = "My Shiny (v1)"
    manufacturer = "Shiny Technologies"
    url = 'http://www.shiny-tech.com/'
    architecture = 'ar71xx'
    usb = True
    usb3 = True
    # ...
```

Instead I only want to have a single attribute to configure USB, the `usb` attribute. So in order to support different USB protocol versions, we need to change how the existing attribute works. My proposal is to change the values that the `usb` attribute can take. 

You need to do the following:
* Define two new constants in `devices.py`, `USB_2` and `USB_3` (e.g., with values `'usb-2'` and `'usb-3'`).
* Change all CGMs (I think this is actually just one, the one that you already changed in this PR, but you should check), which use the `usb` attribute to support the new constants.
* Change the device descriptor validation code to check that the `usb` attribute takes one of the supported values (either `None`, `False`, `USB_2` or `USB_3`).
* Change all existing device descriptors to use `USB_2` instead of `True`.

nodewatcher/core/generator/cgm/devices.py
```diff
@@ -529,6 +529,11 @@ class DeviceBase(object):
     radios = None
     switches = None
     usb = False
+    usb = None
```
kostko, line 532: You don't need multiple of these, this is just the default value and it can be either `None` or `False`. Keep just one (e.g., `usb = None`) and remove the rest.

nodewatcher/core/generator/cgm/devices.py
```diff
@@ -529,6 +529,11 @@ class DeviceBase(object):
...
+    usb = USB_2
+    usb = USB_3
+    USB_2 = 'usb-2'
```
kostko, line 535: Don't define constants in `DeviceBase` class. I would suggest you create a new class, e.g. `USBProtocol` that would hold these two constants. Don't forget a documentation block describing what these constants are.

nodewatcher/modules/devices/glinet/glinet.py
```diff
@@ -1,103 +1,103 @@
...
+    """
+    GL.iNet 6408A v1 device descriptor.
+    """
+
+    identifier = 'gl-inet6408av1'
+    name = "GL.iNet 6408A (v1)"
+    manufacturer = "GL Technologies"
+    url = 'http://www.gl-inet.com/'
+    architecture = 'ar71xx'
+    usb = USB_2
```
kostko, line 16: You need to reference these constants from the correct class. So when you make the change I mentioned above and move the constants out, you should use something like `cgm_devices.USBProtocol.USB_2`.

The same applies elsewhere as well.

kostko (changes requested):

PR #63, opened 2017-09-06
devices: Add CPE version

robimarko:
Add v1 to CPE 210 and 220 models as lately, TP Link is only selling v2 which differs from v1 and currently has no OpenWRT/LEDE support.
Flashing v1 firmware on v2 will cause soft brick


robimarko:
@mitar any chance for merging?

nodewatcher/modules/devices/tplink/cpe.py
```diff
@@ -3,13 +3,13 @@
...
-class TPLinkCPE210(cgm_devices.DeviceBase):
+class TPLinkCPE210v1(cgm_devices.DeviceBase):
     """
-    TP-Link CPE210 device descriptor.
+    TP-Link CPE210 v1 device descriptor.
     """
 
-    identifier = 'tp-cpe210'
-    name = "CPE210"
+    identifier = 'tp-cpe210-v1'
```
kostko, line 11: Existing identifier change requires a database migration as this will break all nodes using the old device identifier.

For an example of how to write a migration, see [this migration](https://github.com/wlanslovenija/nodewatcher/blob/development/nodewatcher/modules/devices/migrations/0001_change_glinet_id.py).

nodewatcher/modules/devices/tplink/cpe.py
```diff
@@ -3,13 +3,13 @@
...
+    name = "CPE210 v1"
```
kostko: Inconsistent naming, check other device descriptors (e.g., it should be in parenthesis, `(v1)`).

nodewatcher/modules/devices/tplink/cpe.py
```diff
@@ -90,16 +90,16 @@ class TPLinkCPE210(cgm_devices.DeviceBase):
...
-class TPLinkCPE220(TPLinkCPE210):
+class TPLinkCPE220v1(TPLinkCPE210v1):
     """
-    TP-Link CPE210 device descriptor.
+    TP-Link CPE210 v1 device descriptor.
     """
 
-    identifier = 'tp-cpe220'
-    name = "CPE220"
+    identifier = 'tp-cpe220-v1'
```
kostko, line 98: Same as above (migration needed).

nodewatcher/modules/devices/tplink/cpe.py
```diff
@@ -90,16 +90,16 @@ class TPLinkCPE210(cgm_devices.DeviceBase):
...
+    name = "CPE220 v1"
```
kostko: Same as above (inconsistent naming).

nodewatcher/modules/devices/tplink/cpe.py
```diff
@@ -90,16 +90,16 @@ class TPLinkCPE210(cgm_devices.DeviceBase):
...
 
 
-class TPLinkCPE510(TPLinkCPE210):
+class TPLinkCPE510(TPLinkCPE210v1):
```
kostko: Is there no versioning for CPE 510? Should we make this v1 just in case?

kostko (changes requested):

nodewatcher/modules/devices/tplink/cpe.py
robimarko: I have not found v2 of CPE510/520 mentioned anywhere, not even on TP Link support site(Usually new versions first pop up there).
I could make them v1 also to future proof them.

nodewatcher/modules/devices/tplink/cpe.py
robimarko: Yes, you are correct.
I will change it today.

nodewatcher/modules/devices/tplink/cpe.py
robimarko: Ok, will make necessary DB migrations today

robimarko: Ok,I made the migrations manually since Django wont detect them.
Only thing that I dont know to configure is dependencies

nodewatcher/modules/devices/migrations/0002_change_cpe_id.py
```diff
@@ -0,0 +1,21 @@
...
+
+from django.db import migrations
+
+from nodewatcher.core.generator.cgm import device_migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cgm', '0016_auto_20160206_2343'),
```
robimarko: Since this is manually made,I dont know what should dependencies be set to

kostko: You should **never** make migrations manually. Use the following command to create an empty migration instead (after removing the manually created migration):
```
python manage.py makemigrations devices --empty -n change_cpe_id
```

And then edit the created file.

kostko: Also see [Django documentation on migrations](https://docs.djangoproject.com/en/1.10/topics/migrations/).

robimarko: Well,I tried to make them but Django refused to detect changes.
I did not know about empty flag.

I have read documentation on migrations already

kostko: Yes, you need to use the `--empty` flag when creating data migrations (as opposed to schema migrations, where changes are detected automatically). This is described in the [section on Data migrations](https://docs.djangoproject.com/en/1.10/topics/migrations/#data-migrations).

robimarko: Thanks for explanation.
This time DB migration should be correct

robimarko:
@kostko This PR should now be ready for merging

nodewatcher/modules/devices/tplink/cpe.py
```diff
@@ -90,22 +90,22 @@ class TPLinkCPE210(cgm_devices.DeviceBase):
     }
 
 
-class TPLinkCPE220(TPLinkCPE210):
+class TPLinkCPE220v1(TPLinkCPE210v1):
     """
-    TP-Link CPE210 device descriptor.
+    TP-Link CPE210 v1 device descriptor.
```
kostko: Fix this description (it was incorrect before as it is for CPE220 not CPE210).

kostko (approved):
Just one minor comment, otherwise looks good.

nodewatcher/modules/devices/tplink/cpe.py
robimarko: Fixed

PR #64, opened 2017-09-07
Bump Django

robimarko:
Bump Django to include security bugfix in 1.10.8.
Nothing else changed.


mitar:
Thanks.

BTW, please make commit messages whole sentences.

mitar:
And I hope you tested that it works. :-)

robimarko:
Yes,from what I tested everything works.
Also I did not expect any issues since only change is security bugfix.

PR #65, opened 2017-11-18
devices: Add multiple Mikrotik devices

robimarko:
This PR is to add multiple Mikrotik devices that I merged into LEDE, RB912, SXT Lites and LHG 5 which were already supported.

Builder for them is here:
https://hub.docker.com/r/otvorenamreza/lede-builder/tags/
Builder docker image name is vd1eff32_master_ar71xx_mikrotik_18_11_17

But again initramfs is not generated.
So current workaround since manually building imagebuilder using make menuconfig results in the same issue is to download https://downloads.lede-project.org/snapshots/targets/ar71xx/mikrotik/lede-ar71xx-mikrotik-vmlinux-initramfs.elf

### DIFF
--- a/nodewatcher/modules/devices/mikrotik/__init__.py
+++ b/nodewatcher/modules/devices/mikrotik/__init__.py
@@ -1,2 +1,2 @@
 # Mikrotik device modules.
-from . import rb2011, rb450, sxt, rbmap, rbwap
+from . import rb2011, rb450, sxt, rbmap, rbwap, rb912
--- a/nodewatcher/modules/devices/mikrotik/__init__.py
+++ b/nodewatcher/modules/devices/mikrotik/__init__.py
@@ -1,2 +1,2 @@
 # Mikrotik device modules.
-from . import rb2011, rb450
+from . import rb2011, rb450, sxt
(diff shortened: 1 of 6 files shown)


kostko:
Are these all supported in our LEDE builders?

robimarko:
I dont understand what do you mean by that?
They are supported in LEDE and I made builder from development branch since no Mikrotik device using 16MB NOR flash is in stable branch.
And Mikrotik made revisions of most devices to only use 16MB NOR and you cant get NAND versions.

kostko:
What I meant is if this requires a non-released version of LEDE? Any timelines on when that will be released? Otherwise I see no problems in merging this, but we should only bump builders to stable LEDE versions.

robimarko:
Yes,it is based on development branch.
There are currently no informations when will new stable version that is based on development branch be realeased.
Currently all stable versions are based of lede-17.01 branch and pretty much all devices and changes except for bugfixes and security issues were not backported.
So stable versions currently lack almost all devices that were added in the last 6 or even more months.

mitar:
You were reporting that generator is not generating all files needed to flash Mikrotik devices. Was this resolved?

robimarko:
No,it is upstream issue.
LEDE imagebuilder also does not generate initramfs files.
That is not really an issue since you can just use LEDE generated snapshot initramfs images

mitar:
Was this reported upstream?

Should then we include those snapshot images and fake it that they were generated?

robimarko:
Yes, someone posted it on forum and mailing list couple of months ago.
It did not catch any traction

I dont think that we should fake them.
We should do the same thing as with Edgerouter X that also requires initramfs image which is also not generated by imagebuilder, just simply use one provided on LEDE website

robimarko:
Any chance of merging?

PR #66, opened 2018-05-05
Fix building

robimarko:
Fix building cause by pulling pip 10.0.1 release as the current one.
That one is broken on all distros and cause docker-compose build to exit with no such option: --allow-all-external error.
--allow-all-external got removed with 10.0.0.
Fixing pip to latest known to work 9.0.3 fixes that.


mitar:
Great catch! Why do we even use `allow-all-external`? And why is CI still failing?

robimarko:
I dont know why do we use it.
It used to enable checksum checking agains pypi when stuff was pulled from third party sources.

I dont know why is CI failing on PostgreSQL version check.
Can you restart the CI build?

mitar:
You cannot restart CI build? You should be able. Login into Travis.

robimarko:
Yeah,I was not logged in.
It failed again on the same error.
It looks like that error is caused by psycopg2 version being too old.
I gotta test that.

I think that --allow-all-external is now default behavior in pip 10 but I gotta test it so we can remove the hard-coded dependency.

Also,during GSOC I gotta get around to updating packages as a lot of them are way out of date.

